PR #9, opened 2021-02-24
versioning and publishing workflow

dianelee217:
## What
  * Added CI/CD workflow for publishing package to pypi
  * Updated versioning
  * Added project classifiers


setup.py
```diff
@@ -25,9 +25,9 @@
 
 setup(
     name="rubicon",
-    version=get_version(),
-    author="The Rubicon Team",
-    author_email="rubicon-developers@capitalone.com",
+    version=__version__,
+    author="Joe Wolfe, Ryan Soley, Diane Lee, CapitalOne",
+    license='Apache License, Version 2.0',
```
mmccarty, line 30: Please add me here. I do intend to contribute.

mmccarty:
Can you also add badges with CI status to the README?

rubicon/version.py
```diff
@@ -0,0 +1,11 @@
+"""
+File containers the version number for the package
+"""
+
+MAJOR               = 0
+MINOR               = 1
+MICRO               = 0
+
+VERSION             = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
```
joe-wolfe21, line 9: @mmccarty - we suggested starting at `v0.1.0`  - do you agree ?

setup.py
```diff
@@ -41,4 +41,27 @@
             "rubicon_experiment = rubicon.intake_rubicon.experiment:ExperimentSource",
         ],
     },
+    classifiers=[
+        # How mature is this project? Common values are
+        #   3 - Alpha
+        #   4 - Beta
+        #   5 - Production/Stable
+        'Development Status :: 4 - Beta',
```
joe-wolfe21, line 49: @mmccarty - we were debating between Beta, or Production/Stable -- what do you think?

mmccarty: I think Beta is fine for now given that we are just now open sourcing and might get some activity that would make things less stable.

joe-wolfe21 (approved):
LGTM

PR #10, opened 2021-02-24
change package name to rubicon-ml to match pypi

dianelee217:
## What
  * Updating package name to match https://pypi.org/project/rubicon-ml/


joe-wolfe21 (approved):
LGTM

PR #15, opened 2021-03-01
GitHub pages workflow

joe-wolfe21:
## What

Adds a github action workflow to automatically build and publish the latest docs. The `build-docs.sh` script used bash syntax for regex matching but I couldn't get it to run on github actions. Looks like the default shell there is `/bin/sh` and even with following the steps to update shell [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-running-a-script-using-bash), it still failed to execute the script. So for now i took out the regex check

## How to Test

Using github actions to test
* https://github.com/capitalone/rubicon/commit/b57132b9bf74aab4df501d341cb9a2fc29b0adbd
* https://github.com/capitalone/rubicon/pull/15/checks?check_run_id=2005494964

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -3,9 +3,12 @@
 # For more details: https://github.com/marketplace/actions/deploy-to-github-pages
 name: Publish docs
 
+# on:
+#   release:
+#     types: [created]
 on:
-  release:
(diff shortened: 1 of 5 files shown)


dianelee217 (approved):
LGTM

Issue #11, opened 2021-02-25
[BUG] Dashboard unable to load experiments which lack metrics/parameters


daviseford:
**Describe the bug**
The Rubicon Dash-powered Dashboard is unable to handle experiments that _do not_ contain metrics and parameters.

**Steps/Code to reproduce bug**

To encounter the bug, run the following python script to generate data.

```python
from rubicon import Rubicon

root_dir = "./rubicon-root" # Replace me with an S3 path if desired
rubicon = Rubicon(persistence="filesystem", root_dir=root_dir)

# Create a project
project = rubicon.create_project("Sample Project")

# Create three empty experiments (note that these do not have metrics/parameters!)
project.log_experiment("Exp1")
project.log_experiment("Exp2")
project.log_experiment("Exp3")
```

Now attempt to load the `rubicon-root` project folder in the Dashboard. I have gotten two varieties of failure:

#### Error 1: Running in a container environment

1. Click on the "Sample Project" button
2. Check the developer console for the following error:

```
 ---dash_renderer.min.js:20 
Object
html: "<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">↵<title>500 Internal Server Error</title>↵<h1>Internal Server Error</h1>↵<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>↵"
message: "Callback error updating grouped-project-explorer.children"--- 
```

#### Error 2: Running the Dashboard locally

I encountered a purely silent failure while running the Dash app locally. 

I could open the Dashboard, but I didn't even see the "Sample Project" button.

**Expected behavior**

Given the following generated data:

```python
from rubicon import Rubicon

root_dir = "./rubicon-root" # Replace me with an S3 path if desired
rubicon = Rubicon(persistence="filesystem", root_dir=root_dir)

# Create a project
project = rubicon.create_project("Sample Project")

# Create three empty experiments (note that these do not have metrics/parameters!)
project.log_experiment("Exp1")
project.log_experiment("Exp2")
project.log_experiment("Exp3")
```

I would expect to be able to load this project in the Dashboard and view my 3 experiments.

**Additional Context**

The following code snippet will generate data that _does_ work with the current Dashboard.

The key difference between this and the code above is that we generate metrics and parameters for each experiment.

```python
from rubicon import Rubicon

root_dir = "./rubicon-root"
rubicon = Rubicon(persistence="filesystem", root_dir=root_dir)

[...]

PR #17, opened 2021-03-01
handle comparison view with no params or metrics

joe-wolfe21:
**closes: #11**

---

## What

If a project's experiments did not have any metrics or parameters logged to them yet, the dashboard could not load the project data properly. This PR handles that case and adds a test against it.

## How to Test

`pytest -m "dashboard_test"`


ryanSoley (approved):
looks good!

PR #20, opened 2021-03-02
auto versioning using git tags

ryanSoley:
## What
  * in `version.py` and `rubicon/_version.py`, use `git` to get the version from the latest tag
    * this needs to be run from the repo with tags cloned too, which means it'll only work for devs who've installed from source (or the CI with the `fetch_depth` maxed out for now, in our case)
  * to deal with that, `setup.py` calls `version.py`'s `_write_version_file` function when the package is bundled
    * this replaces `rubicon/_version.py` with a function that returns a hardcoded string (fetched from the `git` tags in the build environment)
    * this change won't ever need to be committed to the repo, because the current `git` solution will always work for anyone installed from source

## How to Test
  * from the repo's root, run `pip install -e .`, launch a python interpreter, and `import rubicon`
    * `rubicon.__version__` should return the latest version
    * navigate out of the repo and try the same thing - `get_version` will fail since theres no git repo
  * build `rubicon` - `python setup.py sdist bdist_wheel` - and install the wheel file
    * now `rubicon.__version__` should return the latest version from a python interpreter started anywhere

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -17,6 +17,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
--- a/README.md
+++ b/README.md
@@ -134,6 +134,25 @@ Now `pre-commit` will run automatically on git commit and will ensure consistent
 code format throughout the project. You can format without committing via
 `pre-commit run` or skip these checks with `git commit --no-verify`.
 
+## Releasing
+
+Releasing a new version of `rubicon` to PyPi is as simple as drafting a new
+release on GitHub:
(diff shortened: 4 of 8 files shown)


ryanSoley:
okay so I looked into what versioneer was doing and how it solved the issue we were talking about with needing to commit an updated version file to the repo on builds

if you look at [this file](https://github.com/dask/dask/blob/master/dask/_version.py) on a local `dask` install, you'll see a totally different file

so here, `rubicon/_version.py` is where `rubicon.__version__` comes from. `version.py` at the root is for setup only. it'll rewrite `rubicon/_version.py` before its bundled and packaged so that `rubicon/_version.py` is just returning a string rather than getting it from `git`. whenever we're running locally from the repo code, it'll use what's committed here (`git`) at `rubicon/_version.py` which should never change

ryanSoley:
@gforsyth @mmccarty does this make sense? and more importantly, is it a good idea?

@joe-wolfe21 is on board but we wanted another opinion

mmccarty:
I generally recommend [versioneer](https://github.com/python-versioneer/python-versioneer).

gforsyth:
fwiw, I'd just go with versioneer -- it's a bit bloated at times but that's because it handles all of the edge cases.

ryanSoley:
@gforsyth is maintaining versioneer as simple as updating the package, or am I gonna need to make sure that giant file I need to dump in the repo is updated every time the library updates? do people generally pin versioneer?

this probably sounds stupid, but honestly handling all the edge cases is what I dont like about it. I'd rather my code just fail and I know why than having to decipher their stuff to figure out why the latest release got uploaded as version `0+unknown`

I've never used versioneer long term so maybe I'm just making up problems

gforsyth:
> is maintaining versioneer as simple as updating the package, or am I gonna need to make sure that giant file I need to dump in the repo is updated every time the library updates? do people generally pin versioneer?

So you install it the once on your dev machine and the initialize the repo using it, but it puts a standalone file that gets "vendored" in.  So you don't need to update it, or mess with it.

gforsyth:
So, the way the current 'version.py' script is written, it won't alert you if you've installed locally using uncommitted changes, which seems problematic.

ryanSoley:
> So, the way the current 'version.py' script is written, it won't alert you if you've installed locally using uncommitted changes, which seems problematic.

okay yeah that's a good point - the benefits will probably outweigh anything I'm worried about

ryanSoley:
closing to reimplement using versioneer

Issue #22, opened 2021-03-03
automate version string updates


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

for each release, we need to manually update the version strings in `setup.py` and `rubicon/__init__.py`. this is prone to us forgetting to do so.

**Describe the solution you'd like**

we use `git` tags as the source of truth for our versions. the version strings should be populated from this.

**Describe alternatives you've considered**

[versioneer](https://github.com/python-versioneer/python-versioneer)

ryanSoley:
just gonna use versioneer for this

PR #31, opened 2021-03-05
add `versioneer` for versioning

ryanSoley:
closes: #22 

---

## What
  * adds [`versioneer`](https://github.com/python-versioneer/python-versioneer) for versioning
  * cleans up some stuff in our various config files

## How to Test
  * run `python setup.py version` with a local install - you should see version `0.1.2+12.g8731195`
     * the last part will change if I add more commits
     * more on the version style and what each part means [here](https://github.com/python-versioneer/python-versioneer/blob/master/README.md#styles)
  * from an interpreter, run `import rubicon` and `rubicon.__version__` - you should see the same string

## Known Issues
  * when `pre-commit` lints our code through torch, the formatting libraries don't respect their config files at the root of this repo
     * I had to remove the legacy `pre-commit` hooks from torch before I could commit this (which was not intuitive)
     * some possible solutions:
        * just don't use `pre-commit` locally - its only us that have to deal with torch and the CI should catch anything if we forget
        * don't ignore the versioneer files while linting - idk if it causes any problems but versioneer writes files from strings in `versioneer.py` and I dont want those to get messed up


joe-wolfe21 (approved):
Looks good, thanks for getting this setup 👍

Issue #30, opened 2021-03-05
investigate pinned dependencies


ryanSoley:
**Is your enhancement request related to a problem? Please describe**
the conda install for `rubicon-ml` is super slow and takes a few solves to install into a fresh environment. this is almost certainly due to issues with our pinned dependencies

**Describe the solution you'd like**
  * investigate why our pinned dependencies are pinned (`pyarrow`, `pyyaml`, `fsspec`)
  * investigate the dependencies called out by the conda forge bot in the dependency analysis [here](https://github.com/conda-forge/rubicon-ml-feedstock/pull/2) (UI libraries, `pyarrow`, `s3fs`)
  * update and unpin any upper limits we can

ryanSoley:
at a first glance before digging into it, here are a few things I think we'll have to do
  * add the relevant UI libs to the conda recipe (since conda doesn't do extras)
  * unpin `pyyaml` and `fsspec` - I don't think theres any legit reason they're pinned
  * determine if we even need `pyarrow` and `s3fs` - they're optional dependencies of our dependencies

PR #34, opened 2021-03-08
Speed up install

joe-wolfe21:
closes: #30 

---

## What

Unpins the upper bounds on a few dependencies based on the feedback in #30. `s3fs` and `pyarrow` are required as part of the install, unless we want to update to make them optional installs based on how rubicon is used (i.e. s3? -> s3fs). Although for now I think it's fine to leave as is

The install time dropped a couple mins for me

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -2,7 +2,7 @@ name: rubicon-dev
 channels:
   - conda-forge
 dependencies:
-  - python=3.8.0
+  - python=3.8
   - pip
 
   - click>=7.1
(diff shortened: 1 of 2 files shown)


ryanSoley:
> The install time dropped a couple mins for me

like just creating the environment from the file?

ryanSoley:
also, do you think theres any reason not to unpin the python version?

joe-wolfe21:
@ryanSoley yes, from the file directly -- and i'm gonna pin python to 3.8 since that's the highest version we test against right now

ryanSoley (approved):

Issue #32, opened 2021-03-08
Support dataframe.plot


joe-wolfe21:
**Is your feature request related to a problem? Please describe**

The plotting functionality for dataframes raises a `NotImplementedError`. It was dependent on an internal visualization library and will need to be switched to an open source library under the HoloViz ecosystem.

**Describe the solution you'd like**

Reimplement the convenience method dataframe.plot using hvplot

PR #40, opened 2021-03-10
Dataframe plot

joe-wolfe21:
closes: #32 

---

## What

Uses hvplot to adding basic dataframe plotting support

## How to Test

Build the docs and visit the new notebook. I committed with the notebook's output so that the example plots are rendered in the docs to give viewers a quick idea of the plotting functionality

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -23,6 +23,6 @@
 }
 
 /* unknown warning was showing, manually hiding */
-#Default-Plotting-Support .admonition.warning {
+#Visualizing-Logged-Dataframes .admonition.warning {
     display: none;
 }
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -21,3 +21,8 @@
     font-size: 20px;
     margin-top: -14px;
 }
+
+/* unknown warning was showing, manually hiding */
+#Default-Plotting-Support .admonition.warning {
+    display: none;
+}
--- a/rubicon/client/dataframe.py
+++ b/rubicon/client/dataframe.py
@@ -1,5 +1,3 @@
-import sys
-
 from rubicon.client import Base, TagMixin
 from rubicon.exceptions import RubiconException
 
--- a/rubicon/client/dataframe.py
+++ b/rubicon/client/dataframe.py
@@ -3,6 +3,7 @@ import sys
 from rubicon.client import Base, TagMixin
 from rubicon.exceptions import RubiconException
 
+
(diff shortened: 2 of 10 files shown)


rubicon/client/dataframe.py
```diff
@@ -38,9 +41,36 @@ def _get_data(self):
...
+        For customizations, visit:
+        https://hvplot.holoviz.org/user_guide/Customization.html
+
+        Examples
+        --------
+        >>> # Log a line plot
+        >>> dataframe.plot(kind='line', x='Year', y='Number of Subscriptions')
+        """
+        # data is a dask dataframe
+        if sys.modules.get("dask") is not None:
```
ryanSoley, line 65: when would this ever be `None`?

ryanSoley (commented):
I'm getting this warning when building the docs `/Users/nvd215/oss/rubicon/docs/source/plotting-support.ipynb:1: WARNING: Content block expected for the "raw" directive; none found.` and in the rendered output theres a block that says "Data type cannot be displayed:"

there's an empty cell at the end of the notebook - lets get rid of that

also the plot doesnt look great in dark mode, but thats probably more trouble than its worth

ryanSoley:
oh and when i just view the notebook with `jupyter` I dont see the plot - not sure if thats causing any of the issues I mentioned. it renders in the docs when I build them fine

rubicon/client/dataframe.py
joe-wolfe21: i added it as a precaution -- only if we refactored to select either `pandas` or `dask` on install, both are currently installed as part of the dependencies now

joe-wolfe21:
> WARNING: Content block expected for the "raw" directive; none found.

Not sure why this is popping up. Based on our convo, decided to hide the warning of `"Data type cannot be displayed:` within the rendered output

docs/source/_static/custom.css
```diff
@@ -20,4 +20,9 @@
     text-align: right;
     font-size: 20px;
     margin-top: -14px;
+}
+
+/* unknown warning was showing, manually hiding */
+#Default-Plotting-Support .admonition.warning {
+    display: none;
```
ryanSoley, line 27: 🔥

ryanSoley (commented):
last nitpick I promise - I don't love the title of the example. what about "Visualizing Logged Dataframes" just to be more specific

joe-wolfe21:
@ryanSoley thanks for the solid PR feedback, it's ready for another look

ryanSoley (approved):
looks great!

Issue #42, opened 2021-03-12
`percy` is available on `conda-forge`


gforsyth:
If you want to remove the lone `pip` package from the ci `environment.yml` file: https://github.com/conda-forge/percy-feedstock

PR #43, opened 2021-03-12
Use conda incubator action for environment setup

gforsyth:
Unpin Python in environment file to make sure Python version is not
always 3.8

closes: #42

---

## What
  * Uses the `conda-incubator` action for more flexible miniconda setup
  * Unset strict python version in environment file so the version matrix checks all the versions 
  * Add percy to conda instead of using pip

## How to Test
  * I think if the CI passes it works?

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -26,7 +26,7 @@ jobs:
       uses: conda-incubator/setup-miniconda@v2
       with:
         python-version: ${{ matrix.python-version }}
-        miniforge-version: latest
+        miniconda-version: latest
         channels: conda-forge
         channel-priority: true  
         activate-environment: rubicon-dev
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -21,17 +21,14 @@ jobs:
         # fetch all tags so `versioneer` can properly determine current version
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: conda-incubator/setup-miniconda@v2
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Add conda to system path
-      run: |
-        # $CONDA is an environment variable pointing to the root of the miniconda directory
-        echo $CONDA/bin >> $GITHUB_PATH
-    - name: Install dependencies
-      run: |
-        conda env update --file ci/environment.yml --name base
-        pip install setuptools wheel
+        mamba-version: "*"
+        channels: conda-forge, defaults
+        channel-priority: true  
+        activate-environment: rubicon-dev
+        environment-file: ci/environment.yml
     - name: Check formatting with black
       run: |
         black --check .
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -14,7 +14,9 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8]
-
+    defaults:
+      run:
+        shell: bash -l {0}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


gforsyth:
Ok, so the `from __future__ import annotations` is only supported in Python 3.7+, so I've gone ahead and dropped 3.6 by adding a `python_requires` to the `setup.py` and removing it from the CI matrix.

This is _incredibly presumptuous_ of me and there's no need to take this path, I'm just offering it up as an option.
Alternatively, the future annotations are, I believe, only for deferred annotation?  So might not be strictly required if we want to maintain 3.6 support.  That said, the PyData ecosystem is largely following NEP 29 (https://numpy.org/neps/nep-0029-deprecation_policy.html) and 3.6 is no longer in support on that timeline.

ryanSoley:
> That said, the PyData ecosystem is largely following NEP 29 (https://numpy.org/neps/nep-0029-deprecation_policy.html) and 3.6 is no longer in support on that timeline.

thats good enough reason for me to drop it. @joe-wolfe21 agree?

joe-wolfe21:
@gforsyth @ryanSoley yea, i'd say we go for it. Also, good find on the `conda-incubator` action 👍

ci/environment.yml
```diff
@@ -2,7 +2,6 @@ name: rubicon-dev
 channels:
   - conda-forge
 dependencies:
-  - python=3.8
```
joe-wolfe21, line 5: should we follow the strategy of offering different versions of the environment.yaml (`environment-3.7.yaml`, `environment-3.8.yaml`) and an `environment-latest.yaml` for ci? Or leave unpinned entirely?

.github/workflows/test-package.yml
```diff
@@ -13,25 +13,24 @@ jobs:
...
+      run:
+        shell: bash -l {0}
     steps:
     - uses: actions/checkout@v2
       with:
         # fetch all tags so `versioneer` can properly determine current version
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: conda-incubator/setup-miniconda@v2
```
joe-wolfe21, line 26: nice find! this definitely is a cleaner conda setup

ci/environment.yml
gforsyth: With the current setup, the version of Python in a given environment is defined by the actions matrix -- the benefit here being that we could add 3.9 support with just one edit, instead of adding an additional yaml file.
In terms of downstream users, their dependencies will be handled by `conda-forge` / `pip`.

gforsyth: If we want it more explicit, though, that's definitely doable

joe-wolfe21: that makes sense -- I was thinking about when setting up a `rubicon-dev` environment locally

ryanSoley:
@gforsyth there are a couple warnings on the last successful action run. looks like the one is just `miniforge-version` -> `miniconda-version`
> https://github.com/capitalone/rubicon/actions/runs/647423967

gforsyth:
> @gforsyth there are a couple warnings on the last successful action run. looks like the one is just `miniforge-version` -> `miniconda-version`

whoops -- I guess the `miniforge` stuff isn't released yet.  I'll fall back to `miniconda` for now

ryanSoley (approved):
not sure what the deal is with the other `"channels" set on the "environment-file" do not match "channels" set on the action!` warning - they look the same to me

thanks for cleaning this stuff up!

ci/environment.yml
gforsyth: Yeah -- instead of hiding that in a `ci` folder, maybe there can be a `environment-dev.yaml` file located in the repo root?

PR #46, opened 2021-03-17
add logo SVG, remove unused logos

ryanSoley:
## What
  * adds the logo SVG to the repo for safe keeping
  * adds a large PNG logo to be used as the social media preview
    * size/format based on GitHub's recommendations
  * removes the two logos we aren't using
    * they can be re-created from the SVG if need be


joe-wolfe21 (approved):
glad we tracked that svg down

joe-wolfe21 (approved):

PR #51, opened 2021-03-23
add logging to fit and score

joe-wolfe21:
## What

Sets up a way to log params during fitting and metric during scoring


ryanSoley:
nice! definitely makes sense. I think if we wanna follow the logger approach I laid out on the base branch we'll wanna iterate over `step_names, _ = zip(*self.steps)` and pull the specific logger for each one (or just the default for now)

feel free to merge this into the `rubicon-pipeline` branch

Issue #56, opened 2021-03-26
bad JSON writes empty files


ryanSoley:
**Describe the bug**

when logging something that can not be json serialized, rubicon throws an error yet still writes an empty file to the logged item's path. this is problematic because when rubicon tries to read the empty file back in, it throws another json decode error.

**Steps/Code to reproduce bug**

```python
>>> from rubicon import Rubicon
>>>
>>> r = Rubicon(persistence="filesystem", root_dir="./rubicon_root_dir")
>>> p = r.create_project("bug")
>>> e = p.log_experiment()
>>>
>>> param_value = str
>>> type(param_value) 
<class 'type'>
>>>
>>> e.log_parameter(name="fails", value=param_value)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/nvd215/oss/rubicon/rubicon/client/experiment.py", line 141, in log_parameter
    self.repository.create_parameter(parameter, self.project.name, self.id)
  File "/Users/nvd215/oss/rubicon/rubicon/repository/base.py", line 801, in create_parameter
    self._persist_domain(parameter, parameter_metadata_path)
  File "/Users/nvd215/oss/rubicon/rubicon/repository/local.py", line 36, in _persist_domain
    f.write(json.dumps(domain))
  File "/Users/nvd215/oss/rubicon/rubicon/repository/utils/json.py", line 42, in dumps
    return json.dumps(data, cls=DomainJSONEncoder)
  File "/Users/nvd215/miniconda3/envs/rubicon-dev/lib/python3.8/json/__init__.py", line 234, in dumps
    return cls(
  File "/Users/nvd215/miniconda3/envs/rubicon-dev/lib/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/nvd215/miniconda3/envs/rubicon-dev/lib/python3.8/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Users/nvd215/oss/rubicon/rubicon/repository/utils/json.py", line 23, in default
    return super().default(obj)  # pragma: no cover
  File "/Users/nvd215/miniconda3/envs/rubicon-dev/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type type is not JSON serializable
>>>
>>> e.parameters()
Traceback (most recent call last):
[...]

PR #57, opened 2021-03-26
don't write empty files

ryanSoley:
closes: #56 

---

## What
  * checks that json is valid in writes before any filesystem operations take place
    * both S3 and local - memory inherits behavior from local
  * adds tests

## How to Test
  * validate the example in #56 no longer throws the second error


joe-wolfe21 (approved):
nice catch!

PR #58, opened 2021-03-29
filtered logger

ryanSoley:
## What
  * adds our first custom logger - the `FilteredLogger`
    * filter which parameters are logged via `ignore`, `ignore_all`, and `select` args

## How to Test
  * run thru the notebook


rubicon/sklearn/loggers.py
```diff
@@ -1,33 +1,22 @@
-from .base_logger import BaseEstimatorLogger
+from rubicon.exceptions import RubiconException
+from rubicon.sklearn.base_logger import BaseEstimatorLogger
 
 
-class StandardScalerLogger(BaseEstimatorLogger):
-    """
-    This is an example, remove before final PR.
-    """
+class FilteredLogger(BaseEstimatorLogger):
```
joe-wolfe21, line 5: now that we know we're not going to be including a lot of custom loggers, it might make more sense to rename this file to `filter_logger`

joe-wolfe21 (commented):
Great idea with this. Quick thing for the class names `BaseEstimatorLogger` and `FilteredLogger` should probably be either:

`BaseEstimatorLogger` and `FilterEstimatorLogger`
or
`BaseLogger` and `FilterLogger`

Issue #55, opened 2021-03-25
Support filtering out params within experiment table / comparison plot


joe-wolfe21:
**Is your enhancement request related to a problem? Please describe**

When lots of params are logged, the experiment table & corresponding comparison plot becomes a little overwhelming. 

**Describe the solution you'd like**

I'd like the ability to filter out specific params so that the experiment table & comparison plot becomes less cluttered as I explore the data.

**Additional context**

This could be done by filtering out the cols within the experiment table itself since the comparison plot pulls its data from there or by adding filtering solely on the comparison plot itself

PR #60, opened 2021-03-30
Support hiding cols within experiment table and comparison plot

joe-wolfe21:
closes: #55 

---

## What

Adds support to hide columns from within the experiment table. Any col hidden there, will also be removed from the experiment comparison plot. We've seen that plot look extremely cluttered when there are lots of params/metrics so hopefully this is a good solution to declutter easily.

## How to Test

* Check out the dashboard functionality
* Run the dashboard integration tests - note you may have to upgrade `chromedriver`: `brew upgrade chromedriver`

### DIFF
--- a/rubicon/ui/assets/css/project-explorer.css
+++ b/rubicon/ui/assets/css/project-explorer.css
@@ -55,3 +55,39 @@
     background-color: rgba(0, 0, 0, 0);
     margin-left: auto;
 }
+
+.dash-table-container {
+    /* make the action btns line up */
+    margin-top: -40px;
+}
+
+.show-hide {
+    box-shadow: none;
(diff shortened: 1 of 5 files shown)


ryanSoley (commented):
this looks great! I have two comments (that may be out of scope for this)
  * it'd be nice if I could toggle all the checkboxes in the dropdown - like if I have a ton of columns and I only wanna show a few I'm not gonna deselect a hundred checkboxes
    * even better could be show/hide for all metrics/parameters
  * if you reorder the axes on the comparison plot then toggle a column they reset - is this an easy fix?

srilatharanganathan:
@joe-wolfe21 , agree we should open a separate ticket.

PR #62, opened 2021-04-01
Fix dataframe

joe-wolfe21:
## What
* The docs build failed when running through the dataframe visualization notebook. I forgot that `kind` is a valid argument to `hvplot` used to specify the plot type (bar/line/table) so needed to rename `kind` to `df_type` to specify the type of df (pandas/dask)

## How to Test
  * The existing tests should cover all the functionality


ryanSoley (approved):
good catch!

Issue #54, opened 2021-03-25
Configurable pagination for experiment table


joe-wolfe21:
**Is your enhancement request related to a problem? Please describe**

By default the experiment table has a page size of 10 experiments. It'd be nice for this to be configurable on the dashboard or by specifying a CLI argument when spinning up the dashboard so the pagination could be more flexible.

**Describe the solution you'd like**

Configurable page size for the experiment table by either CLI arg or as an option within the dashboard

PR #63, opened 2021-04-01
support specifying page-size on dashboard startup

joe-wolfe21:
closes: #54 

---

## What
  * Add a cli param to specify the page size of the experiment table so it's more flexible based on user perference

## How to Test
  * Try different page sizes 

```
rubicon ui --help
Usage: rubicon ui [OPTIONS]

  Launch the Rubicon Dashboard.

Options:
  --root-dir TEXT           The absolute path to the top level folder holding
                            the Rubicon project(s).  [required]

  -ps, --page-size INTEGER  The number of rows that will be displayed on a
                            page within the experiment table.

  -h, --host TEXT           The host to serve the dashboard on.
  -p, --port TEXT           The port to serve the dashboard on.
  -d, --debug BOOLEAN       Whether or not to run in debug mode.
  --help                    Show this message and exit.
```

### DIFF
--- a/rubicon/cli.py
+++ b/rubicon/cli.py
@@ -20,6 +20,13 @@ def cli():
     required=True,
 )
 @click.option(
+    "--page-size",
+    "-ps",
+    type=click.INT,
+    help="The number of rows that will be displayed on a page within the experiment table.",
+    default=10,
+)
+@click.option(
     "--host", "-h", type=click.STRING, help="The host to serve the dashboard on.", default=None
 )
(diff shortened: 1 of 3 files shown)


ryanSoley:
looks great! only thing is entering less than 1 for page size has some interesting results. we should just check to make sure the value of `page-size` is greater than 0

joe-wolfe21:
@ryanSoley good call, updated 💯

ryanSoley (approved):

PR #68, opened 2021-04-05
don't log empty experiments

ryanSoley:
## What
  * stops the pipeline from logging empty experiments
    * the one with only no score was caused by refitting at the end of the grid search, so it totally makes sense that that experiment didn't have a score since it was only fit. I just set `refit=False` for the sake of the example.
    * the totally empty ones were made because sklearn loves to copy estimators all over the place and sometimes doesn't run anything on the copies.
      * to fix this I moved the experiment instantiation to the `fit` call rather than the initialization
      * I'm pretty sure sklearn will yell at you first if you try to call score without calling fit, but I provided an experiment passthrough on the `score` method in case theres some case you'd wanna do that in

## How to Test
  * run the notebook and validate that all the experiments have data on the dashboard

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -18,6 +18,9 @@ dependencies:
   # required for prefect workflows
   - prefect>=0.12.0
 
+  # required for scikit-learn integration
+  - scikit-learn
+
   # required for UI (including testing)
   - beautifulsoup4>=4.8.2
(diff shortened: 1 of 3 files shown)


joe-wolfe21 (approved):
nice! glad to be rid of those empty experiments

PR #77, opened 2021-04-09
`jupyter-dash` proof-of-concept

ryanSoley:
I've been thinking about how we could get a live example of the dashboard hosted for users for a while now. I saw how the `dask` examples use JupyterLab through binder to show off their task graphs and stuff, so I thought it'd be great if we had a way to run the UI in JupyterLab. Of course we can launch it from lab, but that runs it on a localhost port which may not always be accessible.

Then I came across this blog and thought if we could just use this it'd solve it.
> https://medium.com/plotly/introducing-jupyterdash-811f1f57c02e

I was gonna raise this as an issue but it ended up being super easy to implement, so here it is.

## What
  * uses `jupyter-dash` to instantiate the dashboard app
    * the default option, "external", runs it exactly the same as `dash.Dash` would
    * now there are options for "jupyterlab" and "inline" which display the dashboard in a new JLab window or inline in a notebook respectively

## How to Test
  * run through the added notebook (should we even keep it?)


joe-wolfe21 (commented):
This is a great addition! Makes the dashboard super flexible and great to see the side by side view in lab.

> run through the added notebook (should we even keep it?)

I don't think we should keep this example. Instead we should update the [dashboard docs](https://capitalone.github.io/rubicon/dashboard.html#dashboard) with this updated functionality and list the 3 possible ways of running the dashboard. Each way to run should also list any prereqs required to get it working.

ryanSoley:
> I don't think we should keep this example. 

cool, agreed

> Instead we should update the [dashboard docs](https://capitalone.github.io/rubicon/dashboard.html#dashboard) with this updated functionality and list the 3 possible ways of running the dashboard. Each way to run should also list any prereqs required to get it working.

in this or should I add an issue to the docs milestone?

joe-wolfe21:
> in this or should I add an issue to the docs milestone?

In this PR, it's not a lot to add and will make it visible to users

ryanSoley:
@joe-wolfe21 updated the docs. I think they could use some new pics/gifs, but we may end up tackling that with a video in some of our planned work so I'm not gonna mess with that now

joe-wolfe21 (approved):
Awesome! Nice work

PR #80, opened 2021-04-19
bugfix docs workflow

ryanSoley:
## What
  * so I pretty much rewrote the whole docs environment file
    * unpinned all the doc building stuff - this seems to have fixed the plot viz issue
    * installs the rubicon deps with conda, not pip
  * doc build won't fail on removing non-existent files
  * now when doc build fails, the action will fail

## How to Test
  * check out the docs - I build them off this branch


joe-wolfe21 (approved):
🔥 🚒

joe-wolfe21 (approved):
🥳

PR #81, opened 2021-04-20
remove package-lock that sneaked in somehow

joe-wolfe21:
## What

Removes `package-lock.json` that does not belong in here


ryanSoley (approved):
👋

Issue #82, opened 2021-04-20
UI issue when used with ipython


peroni70:
**Describe the bug**
When using Rubicon in a jupyter notebook (ipython), removing files in the filesystem that store the log project data generates `.ipynb_checkpoints` directories within the root directory. This causes the UI to fail to read and display project and/or experiment data.

**Steps/Code to reproduce bug**
In an environment with jupyter and rubicon installed:

Create a new jupyter notebook, copy the following script into a cell, and run the cell

```
from rubicon import Rubicon
root_dir = "./rubicon-root" 
rubicon = Rubicon(persistence="filesystem", root_dir=root_dir)

project = rubicon.create_project("Sample Project")

project.log_experiment("Exp1")
project.log_experiment("Exp2")
```

From a terminal, navigate to the directory containing the notebook

Run `rubicon ui --root-dir rubicon-root` - you can see the `Sample Project` in the UI with two experiments

Quit the UI

Remove _one_ of the experiment folders, run `rm -r rubicon-root/experiments/{experiment_id}`

Restart the UI with `rubicon ui --root-dir rubicon-root` - you should see the 'Sample Project' but when selected, there will be no experiments displayed. 

**Expected behavior**
When `.ipynb_checkpoints` directories exist within the rubicon root directory, I should still be able to see my projects and experiments in the UI.

PR #84, opened 2021-04-23
82 handle misc dotfiles

joe-wolfe21:
closes: #82 

---

## What

As reported in the bug, if misc files sneak into the `root_dir`, the library and UI return empty data rather than just skipping the misc file. This was noticed by `.ipynb_checkpoints` folders from experimentation in Lab.

This handles misc files under the `root_dir` for both the sync library.

## How to Test

Verify tests included handle edge cases and the bug as described in #82 is no longer an issue

### DIFF
--- a/rubicon/repository/base.py
+++ b/rubicon/repository/base.py
@@ -38,11 +38,13 @@ class BaseRepository:
 
     def _ls_directories_only(self, path):
         """Returns the names of all the directories at path `path`."""
-        return [
+        directories = [
             os.path.join(p.get("name"), "metadata.json")
             for p in self.filesystem.ls(path, detail=True)
             if p.get("type", p.get("StorageClass")).lower() == "directory"
         ]
+
+        return directories
 
     # -------- Projects --------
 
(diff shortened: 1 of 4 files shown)


rubicon/repository/base.py
```diff
@@ -38,12 +38,29 @@ def __init__(self, root_dir, **storage_options):
...
     def _ls_directories_only(self, path):
         """Returns the names of all the directories at path `path`."""
-        return [
+        directories = [
             os.path.join(p.get("name"), "metadata.json")
             for p in self.filesystem.ls(path, detail=True)
             if p.get("type", p.get("StorageClass")).lower() == "directory"
         ]
 
+        # TODO - discuss changing this to ignore directories that have dotfiles in their paths
```
joe-wolfe21, line 47: @ryanSoley pointed this out -- leaving this up for discussion. There are pros and cons to both. One pro I like of relying on `_cat_paths` is that it's extensible to other errors we may encounter in the future rather than only focusing on the dotfile problem

ryanSoley: I'm cool with the `_cat_paths` function. that makes sense to me. it may be useful to throw a warning when it hits an error in there

ryanSoley (approved):
this is great! super glad this won't bug anyone anymore

Issue #74, opened 2021-04-09
test example notebooks in scheduled action


ryanSoley:
**Is your feature request related to a problem? Please describe**

as maintainers, we don't always run the examples enough to know if they're up to date & working

**Describe the solution you'd like**

using a combo of `nbformat` to read and `nbconvert` to execute, I'd like to create a test suite that validates each of the examples in the `notebooks` directory runs against the latest version of rubicon without error. these tests will likely be longer running, especially as we add examples, so we should run them as scheduled actions rather than as part of the PR review pipeline.

> https://nbformat.readthedocs.io/en/latest/api.html

> https://nbconvert.readthedocs.io/en/latest/

maybe with the expectation that we manually run them any time theres a PR with example updates? it looks like it may be possible to only run these tests when certain files change, so thats an option too
> https://github.community/t/is-it-possible-to-run-the-job-only-when-a-specific-file-changes/115484

PR #90, opened 2021-04-28
test example notebooks

ryanSoley:
closes: #74 

---

## What
  * the real additions in this PR are in `tests/notebooks/test_notebooks.py` and `tests/notebooks/utils.py`
    * **I'm currently running the tests as a part of the normal CI since they only add a minute to the run time**
      * **we'll probably have to revisit this as there are more examples**
    * the tests validate that each notebook has been executed in order and that they can run without error
    * the prefect and async notebooks are currently ignored by the run without error tests
      * they could be possible to get running, but probably not in the scope of this - thoughts?
    * the `tests/notebooks/bad_notebooks` directory basically contains tests for the tests
  * the bulk of it is example notebook cleanup
    * each notebook's `root_dir` is read from the environment now so the tests can point somewhere else
    * the concurrent logging notebook didn't need that whole script, just for the concurrent function to be imported
    * moved the viz notebook's data inline

## How to Test
  * run the tests normally `python -m pytest` or just run the notebook tests `python -m pytest -m notebook_test`
  * validate that the tests pass and that they clean up after themselves


joe-wolfe21 (approved):
Looks great! Thanks for updating the docs/readme and for cleaning up the example notebooks in this 👍

PR #92, opened 2021-04-28
pytest warnings quickfix

ryanSoley:
## What
  * fixes the warnings coming from rubicon in the tests
  * the ones that are left are thrown when `prefect` is imported in the tests
    * no idea why - importing `prefect` from an interpreter doesn't throw anything

## How to Test
  * `python -m pytest`


joe-wolfe21 (approved):
🏅

PR #94, opened 2021-05-10
Rubicon ml docs

joe-wolfe21:
## What

Updates the package name to `rubicon_ml` within the docs/notebooks

## How to Test

Build the docs locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ reproducibility for both developers and stakeholders alike. While experimenting,
 the Rubicon dashboard makes it easy to explore, filter, visualize, and share
 recorded work.
 
+p.s. If you're looking for Rubicon, the Java/ObjC Python bridge, visit [this](https://pypi.org/project/rubicon/) instead.
+
 ---
 
 ## Components
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ context necessary for a complete model review and approval.
 Here's a simple example:
 
 ```python
-from rubicon import Rubicon
+from rubicon_ml import Rubicon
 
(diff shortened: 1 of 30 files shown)


README.md
```diff
@@ -78,7 +80,7 @@ experiment.log_metric("accuracy", accuracy)
 Then explore the project by running the dashboard:
 
 ```
-rubicon ui --root-dir /rubicon-root
+rubicon_ml ui --root-dir /rubicon-root
```
ryanSoley, line 83: I didn't change the CLI command - should I?

docs/source/index.rst
```diff
@@ -1,8 +1,8 @@
 .. _rubicon:
 
-=======
-Rubicon
-=======
+===========
+rubicon_ml
```
ryanSoley, line 4: so now we have a PyPi project called `rubicon-ml` and our library, `rubicon_ml`. I'd say that this is where we wanna use the dash name `rubicon-ml` since its the project's name and this is the docs' title. but I can totally see how changing between the two could get confusing. thoughts?

ryanSoley (commented):
you only got the `from rubicon import Rubicon`s in the examples. there are other imports

README.md
joe-wolfe21: probably should have the cli command match the package name -- i could change that in this PR

docs/source/index.rst
joe-wolfe21: haha this is what i was debating too -- i went with the library name since it's the docs for the library but i could see how it can go either way.. maybe we should flip a coin 😂

PR #97, opened 2021-05-19
Update manifest

joe-wolfe21:
## What

Noticed that the assets weren't bundled with the latest package -- updating the manifest to include the name change so those assets get resolved and bundled up.


ryanSoley (approved):
so close to not missing anything

PR #100, opened 2021-06-03
address whitesource vulnerability

ryanSoley:
closes: #96 

---

## What
  * addresses the whitesource vulnerability related to `underscore.js`
    * adds an approved version of `underscore.js` to the docs' source
    * updates the makefile to remove any rendered versions of `underscore.js` from the build and replace them
      * I couldn't figure out where the dependency was actually coming from, so this is the best I got
    * updates `build-docs.sh` to not clone the repo anymore
      * I'm pretty sure this is whats resulting in all the old, leftover files on the `gh-pages` branch

## How to Test
  * I'll start a docs build off this branch and we'll see what happens


joe-wolfe21 (approved):
LGTM

Issue #103, opened 2021-06-08
latest `fsspec` version causes `FileExistsError` if multiple memory filesystems are created


ryanSoley:
**Describe the bug**

when a second memory filesystem is created, it attempts to re-make the root directory. in previous versions of `fsspec` it was a no-op if the directory existed, now it throws a `FileExistsError`

**Steps/Code to reproduce bug**

`python -m pytest` from a new dev environment shows the error in the tests

**Expected behavior**

when a second memory filesystem is created, there should be no error. hopefully `fsspec.mkdir` exposes some kind of `exists_ok` flag

PR #104, opened 2021-06-08
check for `root_dir` before creating

ryanSoley:
closes: #103 

---

## What
  * since `mkdir` throws a `FileExistsError` now, check to see if the file exists first
  * also noticed & fixed another failing test for the artifacts downloads
    * using `os.getcwd()` is probably a lot safer than `"."`

## How to Test
  * validate all the tests run with `fsspec` version `2021.5.0` and `2021.6.0`

### DIFF
--- a/rubicon_ml/client/artifact.py
+++ b/rubicon_ml/client/artifact.py
@@ -1,3 +1,5 @@
+import os
+
 import fsspec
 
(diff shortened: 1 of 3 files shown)


joe-wolfe21 (approved):
🔥 🚒

PR #108, opened 2021-06-11
support date parameters

joe-wolfe21:
## What

A quick user ask was to support date params in addition to datetimes so they could use less granular data.


ryanSoley (approved):
nice!

PR #112, opened 2021-07-22
Patch dashboard so it can render dict values

joe-wolfe21:
## What

When a dict value is logged as a param, the dashboard was failing to render. This bug was reported internally. This bugfix resolves the issue.

## How to Test
 
To reproduce on local:

```py
r = Rubicon(persistence="memory")
p = r.create_project("test")
e = p.log_experiment()
e.log_parameter(name="dict", value={"crashes": "ui"})

db = Dashboard(persistence="memory")
db.run_server()
```


ryanSoley (approved):
looks great! thanks for the quick fix!

Issue #115, opened 2021-08-19
running the dashboard from JupyterHub is troublesome


ryanSoley:
**Describe the bug**

sometimes when running from JupyterHub, `Dash` needs some additional configuration to properly load assets and assign paths. some of these config values are read only and need to be set at creation time, so they cant be updated on an instantiated `Dashboard` object

additionally,  [`JupyterDash`](https://github.com/plotly/jupyter-dash) appears to be unmaintained and a source of some of our dashboard struggles. specifically, the "inline" and "jupyterlab" modes dont work from JupyterHub anymore

**Steps/Code to reproduce bug**

from a notebook running in JupyterHub:
```
from rubicon_ml.ui import Dashboard

Dashboard(persistence="memory", mode="inline").run_server()
```

and 

```
Dashboard(persistence="memory", mode="inline").run_server(mode="inline")
```

navigating to the dashboard URL will show a number of errors

**Expected behavior**

* navigating to the dashboard URL should show the dashboard
* running the dashboard with `mode="inline"` should render the dashboard in the notebook

**Additional context**
steps to fix:
- [x] replace `JupyterDash` with `Dash`
- [x] enable a passthru for kwargs to the `Dash` object
- [x] use an `IFrame` for `run_server(mode="inline")` instead of letting `JupyterDash` handle it

PR #116, opened 2021-08-20
make dashboard proxy configurable

ryanSoley:
closes: #115 
---

## What
* ~exposes `configure_dash_app` which allows users to pass config arguments to `dash.Dash` before the dashboard runs~
* ~changes the dashboard import to `from rubicon_ml.ui.dashboard import Dashboard` to avoid import collisions~
  * ~`configure_dash_app` needs to be run before `Dashboard` is imported as the Dash config is applied at import time~
* rubicon's `Dashboard` class accepts all arguments that can be passed through to `dash.Dash` as `dash_options`
  * thanks @joe-wolfe21 for figuring this out in #117 !
* removes `JupyterDash` in favor of an `IFrame`-based solution for inline viewing in notebooks

## How to Test

```python


dash_options = {"requests_pathname_prefix": "/proxy/8050"}

Dashboard(
    persistence="memory",
    dash_options=dash_options,
).run_server(proxy="http://127.0.0.1:8050::https://my-jupyterhub-server.com/")
```

check out [Dash's API reference](https://dash.plotly.com/reference) for more info on kwargs that can be passed to `dash.Dash` via `configure_dash_app` (like `requests_pathname_prefix`) and to `run_server` (like `proxy`)


joe-wolfe21 (commented):
Looking good, just need a few quick updates:
* remove jupyter-dash as dependency
* update imports and dashboard usage in binder example and notebook examples

joe-wolfe21 (approved):
🥇

PR #117, opened 2021-08-23
restructure how the dash app is created

joe-wolfe21:
## What

Restructures the global dash app creation so that we can support more configuration (added in #116) without having to change the import structure (i.e. so we don't need to separate config call and dashboard creation)

```

dash_options = {
    "requests_pathname_prefix": "/proxy/8050"
}

Dashboard(persistence="memory", dash_options=dash_options).run_server(proxy="http://127.0.0.1:8050::https://my-jupyterhub-server.com/")
```

or with CLI

```
rubicon_ml ui --root-dir /rubicon-root --page-size 20
```


docs/source/dashboard.rst
```diff
@@ -46,23 +51,32 @@ the dashboard inline in a notebook with `run_server_inline`:
...
-a high level grouping of its experiments. This works best if you've been
-using the built in git integration. Select a grouping to view the experiments underneath
-it and sort/filter the data directly within the experiment table. Optionally, compare
-experiments by selecting them within the experiment table and viewing the comparison
-plot below.
+.. code-block:: python
+
+  from rubicon_ml.ui import Dashboard
+
+  Dashboard(persistence="memory", requests_pathname_prefix="/proxy/8050").run_server(proxy="http://127.0.0.1:8050::https://my-jupyterhub-server.com/")
```
ryanSoley, line 60: this is different now isnt it?

rubicon_ml/ui/__init__.py
```diff
@@ -1,3 +1,6 @@
+from rubicon_ml.ui.dashboard import Dashboard
```
ryanSoley, line 1: I think this needs to be after `_check_for_ui_extras`

docs/source/dashboard.rst
joe-wolfe21: ah yea good catch

joe-wolfe21:
@ryanSoley thanks - updated!

PR #119, opened 2021-08-30
update how the dashboard is launched

joe-wolfe21:
## What

The binder example was broken after #116 and #117. This resolves the issue and allows the dashboard to be run inline from the binder example.


ryanSoley (approved):
nice! just launched a binder session from this branch and can confirm its working!

PR #120, opened 2021-09-03
Support creating the dashboard from a rubicon instance

joe-wolfe21:
## What

Currently the dashboard relies on persistence and root-dir information and creates a rubicon instance internally. However, it's common that users who want to display the dashboard already have a properly configured rubicon instance, so it was redundant. This allows just passing the rubicon instance for that case.


ryanSoley (approved):
🔥 🔥 🔥

Issue #124, opened 2021-10-19
standardize getters - experiment


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

rubicon currently offers four different types of "getters"
 
 * get all
 * get one by name
 * get one by ID
 * get many by tags

however, these offerings vary by entity type. experiments, for example, currently have "get many by tags," "get one by ID" and "get all"

**Describe the solution you'd like**

experiments should also have the ability to "get one by name"

specifically, `def experiment(self, id)` should be updated to `def experiment(self, name=None, id=None)` where either name or id is required (but not both)

> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/project.py#L231

there is a caveat that makes this a bit different that #131, #133 and #135 - experiment names are not unique. `experiment`, when given a name, could return more than one experiment. we don't want `experiment` to return multiple experiments, so we should return the most recent experiment with the requested name. perhaps also raise a warning that multiple experiments were found any only one will be returned

then, we should additionally update `def experiments(self, tags=[], qtype="or")` to `def experiments(self, tags=[], qtype="or", name=None)` to account for the case where users intentionally have duplicately named experiments and want all of them

> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/project.py#L263

**Additional context**

we originally only offered "get one by ID" because experiments are logged by ID. so when we have an ID, we can retrieve that experiment directly from the filesystem. for experiment names, we'll need to read all the experiments then filter them in-memory. for this reason, we originally left "get one by name" off of experiments. however, I find myself doing `experiment = [e for e in rubicon.experiments() if e.name == "desired name"][0]` to get the one experiment with a specific name often enough, so we may as well just put it in the library

this change should be fully in the client layer

PR #141, opened 2021-11-01
124 experiment getters

andreafehrman:
closes: #124

---

## What
  * add functionality to get experiment by name and id 

## How to Test
  * test through `test_experiment_by_id` and `test_experiment_by_name`

### DIFF
--- a/rubicon_ml/client/experiment.py
+++ b/rubicon_ml/client/experiment.py
@@ -84,10 +84,10 @@ class Experiment(Base, ArtifactMixin, DataframeMixin, TagMixin):
 
         Parameters
         ----------
-            name : str, optional
-                The name of the metric to get.
-            id : str, optional
-                The id of the metric to get.
+        name : str, optional
+            The name of the metric to get.
+        id : str, optional
+            The id of the metric to get.
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


rubicon_ml/client/project.py
```diff
@@ -228,26 +230,42 @@ def log_experiment(
 
         return Experiment(experiment, self)
 
-    def experiment(self, id):
-        """Get an experiment logged to this project by id.
+    def experiment(self, name=None, id=None):
```
ryanSoley, line 233: I know I mentioned that the order here wouldn't matter, but lets actually switch it so `id` is first for the sake of backwards compatibility. that way users that upgrade won't explicitly have to add the `id` kwarg name to places where they currently have `project.experiments("some-id-string")` 

its still good to explicitly use the `id` kwarg name in the places you changed in the library already just to be explicit, so no need to change those back

rubicon_ml/client/experiment.py
```diff
@@ -84,10 +84,10 @@ def metric(self, name=None, id=None):
 
         Parameters
         ----------
-            name : str, optional
```
ryanSoley, line 87: thanks for fixing all these!

rubicon_ml/client/project.py
```diff
@@ -256,11 +274,13 @@ def _filter_experiments(self, experiments, tags, qtype):
                 for e in experiments
                 if has_tag_requirements(e.tags, tags, qtype)
             ]
+            if name is not None:
```
ryanSoley, line 277: this is only gonna filter by name if there are also tags given to filter by. it should be able to do both individually or together

ryanSoley (approved):
looks great 🔥  thanks for taking care of the changes!

Issue #146, opened 2021-11-04
port experiment table to `viz` submodule


ryanSoley:
**Is your feature request related to a problem? Please describe**

the current dashboard's functionality should be available in the new `viz` submodule

**Describe the solution you'd like**

the experiment table from the existing dashboard should be refactored into a standalone widget in the `viz` submodule. the column toggling should be the only interaction that needs to be ported that isn't built into Dash's data table

all work for this issue should be done treating the `viz-prototypes` branch as the root branch. new development branches should be cut from `viz-prototypes` and all PRs should be made into `viz-prototypes`

**Additional context**

for now, we won't worry about interaction between the experiment table and the single metric correlation plot below. for this reason, we'll leave the columns as un-selectable for now

PR #150, opened 2021-11-08
add experiments table to `viz`

ryanSoley:
closes: #146 

---

## What
  * ports over the experiment table from the existing `ui` submodule
  * adds an example notebook for running the new experiment table

## How to Test
  * if you haven't since the dash updates, create a new dev environment
    * `mamba env create -f environment.yml`
  * run through the example notebook and ensure the new experiment table works the same as the old one
    * https://github.com/capitalone/rubicon-ml/blob/146-experiment-table/notebooks/viz/experiments-table.ipynb

### DIFF
--- a/rubicon_ml/viz/experiments_table.py
+++ b/rubicon_ml/viz/experiments_table.py
@@ -88,6 +88,7 @@ class ExperimentsTable(VizBase):
                 disabled=not self.is_selectable,
                 id="select-all-button",
                 outline=True,
+                style={"display": "none"} if not self.is_selectable else {},
             ),
             dbc.Button(
                 "clear all experiments",
@@ -95,6 +96,7 @@ class ExperimentsTable(VizBase):
                 disabled=not self.is_selectable,
                 id="clear-all-button",
                 outline=True,
+                style={"display": "none"} if not self.is_selectable else {},
             ),
         ]
 
@@ -184,7 +186,11 @@ class ExperimentsTable(VizBase):
         return html.Div(
             [
                 self._to_store(ignore_attributes=["app", "experiments"]),
-                html.Div(header, className="header-row"),
+                html.Div(
+                    header,
+                    className="header-row",
+                    style={"margin-bottom": "-3.5rem"} if not self.is_selectable else {},
+                ),
                 dbc.Row(
                     [
                         *[dbc.Col(button, width="auto") for button in bulk_select_buttons],


rubicon_ml/viz/experiments_table.py
```diff
@@ -0,0 +1,281 @@
...
+            commit_hash = list(commit_hashes)[0]
+
+            self.commit_hash = commit_hash[:7]
+            self.github_url = f"{github_url_root}/tree/{commit_hash}"
+
+        self.app.layout = self._build_frame(self._build_layout())
+
+        _register_callbacks(self.app)
+
+    def _build_layout(self):
```
andreafehrman, line 83: What are the intended functions of these buttons? When I run this in the notebook, it doesn't look like there are any changes being made to the visualization.

rubicon_ml/viz/experiments_table.py
```diff
@@ -0,0 +1,281 @@
...
+            dbc.Button(
+                "clear all experiments",
+                color="primary",
+                disabled=not self.is_selectable,
+                id="clear-all-button",
+                outline=True,
+            ),
+        ]
+
+        experiment_table = dash_table.DataTable(
```
andreafehrman, line 101: When the "show all" button is selected, the "id" column takes up quite a lot of space with empty space on the left side of the ids but the "created_at" column values are blocked. Would it be possible to make the id column smaller and allow for the other columns to be seen more clearly?

rubicon_ml/viz/experiments_table.py
```diff
@@ -0,0 +1,281 @@
...
+
+        commit_hashes = set()
+        show_columns = {"id", "created_at"}
+
+        for experiment in self.experiments:
+            if experiment.commit_hash is not None:
+                commit_hashes.add(experiment.commit_hash)
+                show_columns.add("commit_hash")
+
+            if experiment.model_name is not None:
```
andreafehrman, line 32: In the notebook, the "Model_name" column was included in the table when selecting "show all" despite the values all being empty. Should the empty columns be omitted from "show all" as well?

rubicon_ml/viz/experiments_table.py
ryanSoley: the select and clear all experiment buttons right? they're intentionally disabled in the context of a single experiment table because selecting experiment rows doesn't have any value in the standalone experiment table

when the experiment table is rendered in the dashboard with another widget(s) that takes experiments as an input, the experiment table rows will be selectable and selecting them will update the inputs to the other widget(s)

it may be better to make the buttons disappear in the standalone widget, rather than disable them. or maybe I'll just add some kind of tooltip to explain why they're disabled

rubicon_ml/viz/experiments_table.py
ryanSoley: I'm not seeing the same behavior. maybe it has to do with the size of the browser window you're viewing it in?

rubicon_ml/viz/experiments_table.py
ryanSoley: yeah I was debating that one as well. when the widget loads, `model_name` is hidden (as well as any other totally null column). I originally had show all only show the ones that weren't null, but then I thought "that isn't really _all_, is it?" so I decided that the show all button would literally be all regardless of the values. I'm definitely not tied to that idea if you think it'd make more sense to only show the non-null ones

rubicon_ml/viz/experiments_table.py
ryanSoley: @andreafehrman good call - I ended up hiding them instead of just disabling them and I think it looks a lot better and less confusing

rubicon_ml/viz/experiments_table.py
andreafehrman: Looks good now! (Maybe because the buttons were disabled, it fixed itself) Thanks!

rubicon_ml/viz/experiments_table.py
andreafehrman: Ok that makes sense! I think it should be fine to just leave. We could also remove the null columns from the table and just have the `toggle columns` drop down have those selected with (N/A) or (null) next to them? For example, when the option `model_name` has no values, the toggle drop down could have it checked when selecting show all but change it to `model_name (N/A)`. Or we could explicitly say "null" in each of the cells for model_name so that it's more readable and the users can understand that it's not an error. What do you think would be best?

ryanSoley: for now, I'm just gonna leave it as is. I just double checked what the current dashboard does, and it shows empty ones. so we'll leave them the same for now. those are all good ideas though, I'd like to get some user feedback to see which they'd prefer

Issue #137, opened 2021-10-26
standardize getters - artifact


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

rubicon currently offers four different types of "getters"
 
 * get all
 * get one by name
 * get one by ID
 * get many by tags

however, these offerings vary by entity type. artifacts, for example, currently have "get all"

**Describe the solution you'd like**

artifacts should also have the ability to "get one by id" and "get one by name" (they should probably be able to be tagged too, but that's way out of the scope of this ticket. we can circle back to that)

specifically, `def artifact(self, name=None, id=None)` should be created where either name or id is required (but not both) and added to the `ArtifactMixin`

> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/mixin.py#L32

there is already a [`get_artifact_metadata` method on the base repository](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/repository/base.py#L270), so that can be used to construct the `artifact` method on the client in a similar way to the last few tickets like this one

there is a caveat that makes this a bit different that #131, #133 and #135 - artifact names are not unique. `artifact`, when given a name, could return more than one artifact. we don't want `artifact` to return multiple artifacts, so we should return the most recent artifact with the requested name. perhaps also raise a warning that multiple artifacts were found any only one will be returned

then, we should additionally update `def artifacts(self)` to `def artifacts(self, name=None)` to account for the case where users intentionally have duplicately named artifacts and want all of them

> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/mixin.py#L191

**Additional context**

we originally only offered "get one by ID" because artifacts are logged by ID. so when we have an ID, we can retrieve that artifact directly from the filesystem. for artifact names, we'll need to read all the artifacts then filter them in-memory. for this reason, we originally left "get one by name" off of artifacts. however, I find myself doing `artifact = [e for e in rubicon.artifacts() if e.name == "desired name"][0]` to get the one artifact with a specific name often enough, so we may as well just put it in the library

this change should be fully in the client layer

PR #152, opened 2021-11-09
137 artifact getters

andreafehrman:
closes: #137

---

## What
  * added `artifact()` function to get an artifact by name or id 
  * added functionality to `artifacts()` function in order to get artifacts by name

## How to Test
  * use tests `test_artifacts_by_name`, `test_artifact_by_name`, `test_artifacts_by_id` and `test_artifact_name_not_found_error`.

### DIFF
--- a/rubicon_ml/client/mixin.py
+++ b/rubicon_ml/client/mixin.py
@@ -200,13 +200,16 @@ class ArtifactMixin(MultiParentMixin):
         project_name, experiment_id = self._get_parent_identifiers()
 
         if name is not None:
-            self._artifacts = [
+            artifacts = [
                 client.Artifact(a, self)
                 for a in self.repository.get_artifacts_metadata(
                     project_name, experiment_id=experiment_id
                 )
                 if a.name == name
             ]
+            if len(artifacts) == 0:
+                raise RubiconException(f"No artifacts found with name {name}.")
+            self._artifacts = artifacts
         else:
             self._artifacts = [
                 client.Artifact(a, self)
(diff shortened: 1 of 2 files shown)


rubicon_ml/client/mixin.py
```diff
@@ -198,15 +199,59 @@ def artifacts(self):
...
+        rubicon.client.Artifact
+            The artifact logged to this project with id `id` or name 'name'.
+        """
+        if (name is None and id is None) or (name is not None and id is not None):
+            raise ValueError("`name` OR `id` required.")
+
+        if name is not None:
+            artifacts = [a for a in self.artifacts() if a.name == name]
+            if len(artifacts) == 0:
+                raise RubiconException(f"No artifact found with name {name}.")
```
ryanSoley, line 241: you could move the exception into the `artifacts` function then use `self.artifacts(name=name)` here instead directly of just `self.artifacts()`. then the exception would be thrown from both functions and not be duplicated

rubicon_ml/client/mixin.py
```diff
@@ -198,15 +199,58 @@ def artifacts(self):
...
+        Returns
+        -------
+        rubicon.client.Artifact
+            The artifact logged to this project with id `id` or name 'name'.
+        """
+        if (name is None and id is None) or (name is not None and id is not None):
+            raise ValueError("`name` OR `id` required.")
+
+        if name is not None:
+            artifacts = [a for a in self.artifacts() if a.name == name]
```
ryanSoley, line 241:
```suggestion
            artifacts = self.artifacts(name=name)
```

ryanSoley (approved):
looks great!

Issue #139, opened 2021-10-28
add name parameter - dataframe


shania-m:
Like experiments, projects, and metrics, dataframes should also have a name attribute as well. 
- [ ] add name to the domain https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/domain/dataframe.py
- [ ] add name to log_dataframe in the client https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/mixin.py#L228
- [ ] add a name property to the client https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/dataframe.py#L88

Issue #136, opened 2021-10-26
standardize getters - dataframe


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

rubicon currently offers four different types of "getters"
 
 * get all
 * get one by name
 * get one by ID
 * get many by tags

however, these offerings vary by entity type. dataframes, for example, currently have "get many by tags" and "get all"

**Describe the solution you'd like**

dataframes should also have the ability to "get one by id" and "get one by name"

specifically, `def dataframe(self, name=None, id=None)` should be created where either name or id is required (but not both) and added to the `DataframeMixin`

> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/mixin.py#L225

there is already a [`get_dataframe_metadata` method on the base repository](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/repository/base.py#L473), so that can be used to construct the `dataframe` method on the client in a similar way to the last few tickets like this one

there is a caveat that makes this a bit different that #131, #133 and #135 - dataframe names are not unique. `dataframe`, when given a name, could return more than one dataframe. we don't want `dataframe` to return multiple dataframes, so we should return the most recent dataframe with the requested name. perhaps also raise a warning that multiple dataframes were found any only one will be returned

then, we should additionally update `def dataframes(self, tags=[], qtype="or")` to `def dataframes(self, tags=[], qtype="or", name=None)` to account for the case where users intentionally have duplicately named dataframes and want all of them

> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/mixin.py#L268

**Additional context**

we originally only offered "get one by ID" because dataframes are logged by ID. so when we have an ID, we can retrieve that dataframe directly from the filesystem. for dataframe names, we'll need to read all the dataframes then filter them in-memory. for this reason, we originally left "get one by name" off of dataframes. however, I find myself doing `dataframe = [e for e in rubicon.dataframes() if e.name == "desired name"][0]` to get the one dataframe with a specific name often enough, so we may as well just put it in the library

this change should be fully in the client layer

ryanSoley:
closed by #153

PR #153, opened 2021-11-09
Name property to Dataframe and Dataframe getters by name or id

shania-m:
closes: #139 #136

---

## What
  * added name attribute to client, domain dataframe
  * and name parameters to domain dataframes() and project dataframes()
  * added name and id getters for client dataframe()

## How to Test
  * python -m pytest

### DIFF
--- a/rubicon_ml/client/mixin.py
+++ b/rubicon_ml/client/mixin.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import warnings
 from datetime import datetime
 
 import fsspec
@@ -225,7 +226,7 @@ class ArtifactMixin(MultiParentMixin):
 class DataframeMixin(MultiParentMixin):
     """Adds dataframe support to a client object."""
 
-    def log_dataframe(self, df, description=None, tags=[]):
+    def log_dataframe(self, df, description=None, name=None, tags=[]):
         """Log a dataframe to this client object.
 
         Parameters
@@ -243,7 +244,12 @@ class DataframeMixin(MultiParentMixin):
         rubicon.client.Dataframe
             The new dataframe.
         """
-        dataframe = domain.Dataframe(parent_id=self._domain.id, description=description, tags=tags)
+        dataframe = domain.Dataframe(
+            parent_id=self._domain.id,
+            description=description,
+            name=name,
+            tags=tags,
+        )
 
         project_name, experiment_id = self._get_parent_identifiers()
         self.repository.create_dataframe(dataframe, df, project_name, experiment_id=experiment_id)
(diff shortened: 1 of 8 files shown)


rubicon_ml/client/mixin.py
```diff
@@ -293,6 +299,33 @@ def dataframes(self, tags=[], qtype="or"):
 
         return self._dataframes
 
+    def dataframe(self, name=None, id=None):
+
+        if (name is None and id is None) or (name is not None and id is not None):
+            print("hello")
+            print(name)
+            print(id)
```
ryanSoley, line 307:
```suggestion
```

tests/unit/client/test_mixin_client.py
```diff
@@ -172,6 +175,30 @@ def test_dataframes(project_client, test_dataframe):
...
+    project = project_client
+    df = test_dataframe
+    dataframe_a = DataframeMixin.log_dataframe(project, df, name="test_df")
+
+    dataframe_b = DataframeMixin.dataframe(project, name="test_df")
+
+    # assert len(dataframes) == 1
+    assert dataframe_a.name == dataframe_b.name
+
+    # assert dataframe_a.id in [d.id for d in dataframes]
```
ryanSoley, line 188:
```suggestion
    assert dataframe_a.name == dataframe_b.name
```

rubicon_ml/client/mixin.py
```diff
@@ -326,17 +333,61 @@ def dataframes(self, tags=[], qtype="or"):
...
+            ]
         self._filter_dataframes(dataframes, tags, qtype)
 
         return self._dataframes
 
+    def dataframe(self, name=None, id=None):
+        """
+        Get the dataframe logged to this client object.
+
+        Parameters
```
andreafehrman, line 359: These params say artifact instead of dataframe.

rubicon_ml/client/mixin.py
```diff
@@ -326,17 +333,61 @@ def dataframes(self, tags=[], qtype="or"):
...
+        -------
+        rubicon.client.Dataframe
+            The dataframe logged to this project with id `id` or name 'name'.
+        """
+        if (name is None and id is None) or (name is not None and id is not None):
+            raise ValueError("`name` OR `id` required.")
+
+        elif name is not None:
+            dataframes = self.dataframes(name=name)
+            if len(dataframes) == 0:
```
andreafehrman, line 380: I think changing the location of this Rubicon exception to the `dataframes` function would probably be better so it can then throw and exception for the circumstance where the name passed to both `dataframe` and `dataframes` don't exist. I did that in the `artifacts` function as well for reference.

rubicon_ml/client/mixin.py
```diff
@@ -326,17 +333,61 @@ def dataframes(self, tags=[], qtype="or"):
...
+                raise RubiconException(f"No dataframe found with name {name}.")
+            elif len(dataframes) > 1:
+                warnings.warn(
+                    f"Multiple dataframes found with name {name}."
+                    " Returning most recently logged."
+                )
+            dataframes = dataframes[-1]
+        else:
+            project_name, experiment_id = self._get_parent_identifiers()
+            dataframes = self.repository.get_dataframe_metadata(
```
andreafehrman, line 390: I think you might need to call this with `client.Dataframe(self.repository.... , self)` to make sure its returning a Dataframe object. Also I realized that you're calling your variables 'dataframes' but maybe changing it to `dataframe` when it's just one `dataframe` object would make it more readable.

tests/unit/client/test_mixin_client.py
```diff
@@ -213,6 +217,58 @@ def test_dataframes(project_client, test_dataframe):
     assert dataframe_b.id in [d.id for d in dataframes]
 
 
+def test_dataframes_by_name(project_client, test_dataframe):
```
andreafehrman, line 220: Maybe logging one more dataframe with a different name would be good to add to this test to make sure it's grabbing only the ones with the name provided.

rubicon_ml/client/mixin.py
shania-m: done thanks !

andreafehrman (approved):
looks good!!

PR #156, opened 2021-11-19
use `fsspec` to create path for local dataframes

ryanSoley:
## What
  * the `pathlib` `mkdir` call just wasn't making any dirs
  * uses `fsspec`'s `mkdir` instead of `pathlib`

## How to Test
  * log a dataframe to the local filesystem and validate that it works
  * I tested this against S3 to validate that it did not break S3 dataframe logging


shania-m (approved):
LGTM

PR #158, opened 2021-11-29
use mamba in all workflows

ryanSoley:
## What
  * using `conda-incubator/setup-miniconda@v2` with `mamba` worked great in #130, so let's use it for everything

## How to Test
  * release and see what happens


joe-wolfe21 (approved):
🚀

PR #164, opened 2022-01-04
whitesource "fix"

ryanSoley:
## What
  * temporarily drop python 3.7 support


ryanSoley:
since numpy has dropped support for python 3.7 per NEP-29, we might as well too if it'll make whitesource happy
> https://numpy.org/neps/nep-0029-deprecation_policy.html

PR #93, opened 2021-05-10
rename package to `rubicon_ml`

ryanSoley:
closes: #85 
---

## What
  * renames the package from `rubicon` to `rubicon_ml`
  * original `rubicon` imports are still available and warn users to switch their imports
    * these original `rubicon` imports will be removed in the next version after the rename to give users a chance to switch
  * updates the tests to reference `rubicon_ml`

## To Do
- [x] rename repo to rubicon-ml
- [x] update links to repo (main README, dashboard, contributing docs, docs config for notebooks links)
- [x] new logo

## How to Test
  * first, reinstall `rubicon_ml` locally: `pip install -e ".[all]"`
    * you may just wanna recreate the whole environment for consistency's sake
  * run the tests: `python -m pytest`
  * run the example notebooks

PR #169, opened 2022-01-20
remove old `rubicon` module

ryanSoley:
## What
  * removes the old `rubicon` module that was deprecated in #93 

## How to Test
  * if the CI passes it should be good to go


shania-m (approved):
LGTM

Issue #172, opened 2022-02-01
`publish` should be a standalone function


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

to publish a `rubicon_ml` experiment, users currently need an instantiated `Rubicon` object. then they pass it a project that must be from the same filesystem the `Rubicon` object was instantiated with. this workflow is cumbersome and error prone. also, a goal of `intake` is to gather disparate data sources - so it doesn't make sense to limit the experiments in a catalog to a single filesystem

**Describe the solution you'd like**

`publish` should be its own standalone method, not attached to an instance of a `Rubicon` object

PR #175, opened 2022-02-01
better test config & dev docs

ryanSoley:
## What
  * I noticed the test config and dev docs were a bit outdated while doing #172 
    * removes the `dashboard_test` mark
      * dashboard tests don't require chromium for the time being, so they are no longer gated
    * concise, descriptive names for remaining marks
    * update docs for remaining marks

## How to Test
  * `pytest` should run the notebook tests, but not the file writing tests
  * `pytest -m "not run_notebooks and not write_files"` should not run the notebook tests or the file writing tests
  * `pytest -m "write_files"` should run the notebooks tests and the file writing tests
    * the file writing tests will fail since there is no `--s3-path` provided - this is fine

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,27 +135,25 @@ tests/integration`. Or by simply running `pytest` to execute all of them.
 **Note**: some integration tests are intentionally `marked` to control when they
 are run (i.e. not during CICD). These tests include:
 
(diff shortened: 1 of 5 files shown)


shania-m (approved):
LGTM 🎊

Issue #173, opened 2022-02-01
don't auto-publish projects with `intake`


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

currently, a project is automatically published with every set of experiments. this is redundant as each experiment has an attribute to access the project associated with it. furthermore, after #172 there may not be a single project to automatically publish

**Describe the solution you'd like**

projects are no longer automatically published with experiments

**Additional context**

it may be necessary to add a new way to explicitly publish projects, but I'm not sure if users would need this

PR #181, opened 2022-02-15
remove project source from `intake_rubicon`

ryanSoley:
closes: #173
---

## What
  * no longer represent full projects in the intake catalogs
    * publishing projects is redundant - each experiment has a project attribute
    * publishing a full project can still be achieved with `publish(project.experimets())`

## How to Test
  * validate `pytest` passes
  * run through the quick look notebook and validate that there is no longer a project source in the written yaml, only experiments


shania-m (approved):
LGTM

PR #183, opened 2022-02-16
`publish` docs

ryanSoley:
## What
  * docs need the `intake` package in their env now (and `black` for some reason)


shania-m (approved):
LGTM

Issue #182, opened 2022-02-16
Match Rubicon Pipeline signature with Scikit-learn Pipeline signature


shania-m:
**Is your enhancement request related to a problem? Please describe**
The [Scikit-learn Pipeline constructor](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html#sklearn.pipeline.Pipeline) initializes the a Pipeline object with `def __init__(self, steps, *, memory=None, verbose=False)` while Rubicon's [Scikit-Learn Pipeline's](https://github.com/capitalone/rubicon-ml/blob/7f0f0e237ce5a3a208cb674baf7ef6133fd1e5a6/rubicon_ml/sklearn/pipeline.py#L50) defines all extra arguments as `kwargs`.

**Describe the solution you'd like**
Rubicon should replace the current implementation  `def __init__(self, project, steps, user_defined_loggers={}, experiment_kwargs={"name": "RubiconPipeline experiment"}, **kwargs)` with `def __init__(self,project,steps,user_defined_loggers={},memory=None, verbose=False))` so that the Scikit-learn api's behavior is consistent with Rubicon's behavior.

Issue #178, opened 2022-02-07
Update Rubicon Scikit-learn pipelines to support make_pipeline


shania-m:
**Is your enhancement request related to a problem? Please describe**
The scikit-learn api supports a method [make_pipeline](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.make_pipeline.html) as a wrapper around the `Pipeline` constructor. This shorthand does not require the naming of estimators in the pipeline. I would like the Rubicon Pipeline to support the instantiation of pipelines without the explicit naming of the estimators. The steps assign their names as their types in lowercase automatically.
**Describe the solution you'd like**

This solution would allow user calling make_pipeline without naming the steps like so:
````
anova_pipe = make_pipeline(anova_filter,clf)
anova_pipe.fit(X_train, y_train)
print(anova_pipe)
````

Ouput: 
````
RubiconPipeline(project=<rubicon_ml.client.project.Project object at 0x1a4cb8970>,steps=[('selectkbest', SelectKBest(k=3)), ('linearsvc', LinearSVC())])
````

**Comment by ryanSoley** [comment]
I would add that we should look into providing our user-defined loggers to `make_pipeline` as well. something like
```
anova_pipe = make_pipeline(
    steps=(anova_filter, AnovalLoggerCls), (clf, ClassifierLoggerCls)),
)
```

PR #185, opened 2022-02-23
178 Make Pipeline, remove **kwargs from pipeline constructor 

shania-m:
closes: #182 #178

---

## What
  * Added `make_pipeline` method to support constructing unnamed sklearn pipelines 
  * Changed signature for the `pipeline` to  `def __init__( self, project,steps,user_defined_loggers={},experiment_kwargs{"name": "RubiconPipeline experiment"},memory=None,verbose=False):`

## How to Test
  * ensure that all the tests still run with `pytest`
  * ensure that the sklearn integration notebook still runs without errors


rubicon_ml/sklearn/pipeline.py
```diff
@@ -53,15 +54,16 @@ def __init__(
         steps,
         user_defined_loggers={},
         experiment_kwargs={"name": "RubiconPipeline experiment"},
-        **kwargs
+        memory=None,
+        verbose=False,
```
shania-m, line 58: FYI @ryanSoley `black` on pre-commit keeps adding a trailing comma here

rubicon_ml/sklearn/pipeline.py
```diff
@@ -135,3 +137,80 @@ def get_estimator_logger(self, step_name=None, estimator=None):
...
+
+
+def make_pipeline(
+    project,
+    *steps,
+    experiment_kwargs={"name": "RubiconPipeline experiment"},
+    memory=None,
+    verbose=False
+):
+    """Wrapper around RubicionPipeline(). Does not require naming for estimators
```
ryanSoley, line 149: docstring needs a little clean up - mainly removing `user_defined_loggers` and `kwargs` and replacing them with `memory` and `verbose`. you can just copy the docstrings from sklearn over to this (except the `:class:Pipeline` part - i dont think our docs build understands that syntax)
> https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/pipeline.py#L807

ryanSoley (commented):
looking good! thanks for being thorough with the tests! just left a few nitpicky things with the docstrings

other than that, last thing is to add `make_pipeline` to the library ref in the docs. I think itll be something like `.. autofunction:: rubicon_ml.sklearn.pipeline.make_pipeline`
> https://github.com/capitalone/rubicon-ml/blob/main/docs/source/library_reference.rst#sklearn

ryanSoley (approved):

PR #189, opened 2022-03-01
support `NoneType` values in correlation plot

ryanSoley:
## What
  * fixes a bug where experiments that don't have the exact same metrics and parameters logged to them cause the metric correlation plot to error
    * use `.get` to index param/metric dicts to allow for `None`s
    * merges logic for boolean and string data cleaning since they do the same thing


shania-m (approved):
LGTM

Issue #187, opened 2022-02-28
Update Rubicon Scikit-learn pipelines to support score_Samples 


shania-m:
**Is your feature request related to a problem? Please describe**
From #179 :

> `Score_samples` on the other hand is used to score individual scores across samples. This could be sum or the mean of all these scores are used to calculate score() for many estimators; such as the `FactorAnalysis` Estimator, `PCA` estimator, `BayesianGaussianMixture` estimator, and `HalvingGridSearchCV` estimator. Score_samples can also be used for density estimation. Scikit-learn examples show score_samples being used in [Density Estimation](https://scikit-learn.org/stable/modules/density.html?highlight=score_samples), [Density Estimation for a Gaussian Mixture](https://scikit-learn.org/stable/auto_examples/mixture/plot_gmm_pdf.html?highlight=score_samples), [Kernel Density Estimation for Species Distributions ](https://scikit-learn.org/stable/auto_examples/neighbors/plot_species_kde.html?highlight=score_samples), and [Simple 1D Kernel Density Estimation](https://scikit-learn.org/stable/auto_examples/neighbors/plot_kde_1d.html?highlight=score_samples).
> 
> Since `score_samples()` can be explicitly used to density estimation and is used to calculate `scores()`, Rubicon should support `score_samples()` and logging. Similar to the solution proposed in #176, when a user calls `score_samples()`, a new experiment should be opened unless a user specifies which experiment to log to. 

**Describe the solution you'd like**

Rubicon users should be able to call `score_samples()` independently of `score()` and have a score_sample metric logged to Rubicon:

```
project = rubicon.get_or_create_project("Rubicon Pipeline Example")
X = np.array([[-1, -1], [-2, -1], [-3, -2], [1, 1], [2, 1], [3, 2]])
pipe = RubiconPipeline(project,["kde",KernelDensity(kernel='gaussian', bandwidth=0.2).fit(X)])
pipe.score_samples(X)
```

This should result in rubicon logging `score_samples`  to metrics

PR #192, opened 2022-03-03
187 score samples

shania-m:
closes: #187

---

## What
 * `score_samples` added to `rubicon_ml.sklearn.pipeline` and logs an experiment with a `score_samples` metric. Score_Samples are used across samples rather than an aggregate score that would be measured in `score()`. 

>  Score_samples on the other hand is used to score individual scores across samples. This could be sum or the mean of all these scores are used to calculate score() for many estimators; such as the `FactorAnalysis` Estimator, `PCA` estimator,`BayesianGaussianMixture` estimator, and `HalvingGridSearchCV` estimator.

## How to Test
  * pytest -m should pass all tests

### DIFF
--- a/rubicon_ml/sklearn/pipeline.py
+++ b/rubicon_ml/sklearn/pipeline.py
@@ -1,5 +1,6 @@
-from sklearn.pipeline import Pipeline
+from sklearn.pipeline import Pipeline, _name_estimators
 
+from rubicon_ml.client.project import Project
 from rubicon_ml.sklearn.estimator_logger import EstimatorLogger
 from rubicon_ml.sklearn.utils import log_parameter_with_warning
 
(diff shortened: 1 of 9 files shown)


rubicon_ml/sklearn/pipeline.py
```diff
@@ -131,6 +132,34 @@ def score(self, X, y=None, sample_weight=None):
...
+            If not None, this argument is passed as sample_weight keyword argument to the
+            score method of the final estimator.
+        """
+        if experiment is None:
+            experiment = self.project.log_experiment(**self.experiment_kwargs)
+        self.experiment = experiment
+        score_samples = super().score_samples(X)
+
+        if type(score_samples) is ndarray:
+            score_samples = super().score_samples(X).tolist()
```
ryanSoley, line 155: what's going on here? why are we calling `score_samples` twice in this case?

shania-m: good catch, that should just say  `score_samples = score_samples.tolist()` and that's because rubicon can't log ndarrays

rubicon_ml/sklearn/pipeline.py
```diff
@@ -131,6 +131,35 @@ def score(self, X, y=None, sample_weight=None):
...
+        X : iterable
+            Data to predict on. Must fulfill input requirements of first step of the pipeline.
+        y : iterable, optional
+            Targets used for scoring. Must fulfill label requirements for all steps of the pipeline.
+        sample_weight : list, optional
+            If not None, this argument is passed as sample_weight keyword argument to the
+            score method of the final estimator.
+        """
+        if experiment is None:
+            experiment = self.project.log_experiment(**self.experiment_kwargs)
```
ryanSoley, line 149: I think we should follow the same logic as `score` and check `self.experiment` too

Issue #177, opened 2022-02-07
Allow array slicing into Rubicon Scikit-learn pipelines


shania-m:
**Is your enhancement request related to a problem? Please describe**

When slicing a Scikit-learn pipeline, it will return a new slice(copy) of the steps defined in the original pipeline. Rubicon Scikit-learn pipelines support indexing but not slicing. When slicing Rubicon Scikit-learn  pipeline, the Scikit-learn  API returns an Scikit-learn pipeline, which is no a Rubicon Pipeline, causing the program to error: `RubiconPipeline.__init__() missing 1 required positional argument: 'steps'`.

**Describe the solution you'd like**

A user should be able to slice the steps of a RubiconPipeline:
````
log_scale_transformer = RubiconPipeline(
    project,
    [('FunctionTransformer', FunctionTransformer(np.log, validate=False)), ('svc', StandardScaler())],
)
log_scale_transformer[:-1]
````
This should return a New Pipeline with just all the elements except the last one ` Pipeline(steps=[('FunctionTransformer', FunctionTransformer(func=<ufunc 'log'>))])`.

PR #194, opened 2022-03-07
177 pipeline slicing

shania-m:
closes: #177

---

## What
  * Enables slicing on Rubicon Pipelines. The `__get_item__()` method from `sklearn.pipeline.Pipeline` is replaced in `Rubicon_ml.sklearn.Pipeline `with the same functionality except it now passes in the correct arguments to instantiate a `Rubicon_ml.sklearn.Pipeline` class

## How to Test
  * Running `pytest` should not introduce any failures

### DIFF
--- a/tests/unit/sklearn/test_pipeline.py
+++ b/tests/unit/sklearn/test_pipeline.py
@@ -1,9 +1,12 @@
+from tempfile import mkdtemp
 from unittest.mock import patch
+
+from pytest import raises
 
 from rubicon_ml.sklearn import RubiconPipeline
 from rubicon_ml.sklearn.estimator_logger import EstimatorLogger
 from rubicon_ml.sklearn.filter_estimator_logger import FilterEstimatorLogger
-from rubicon_ml.sklearn.pipeline import Pipeline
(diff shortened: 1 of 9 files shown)


tests/unit/sklearn/test_pipeline.py
```diff
@@ -142,14 +142,38 @@ def test_pipeline_memory_verbose(project_client, fake_estimator_cls):
...
+
+    pipeline = RubiconPipeline(project, steps, user_defined_loggers, memory=cachedir)
+    assert pipeline[1:].steps == steps[1:]
+    assert pipeline[1:].user_defined_loggers == {"est1": est1_logger}
+    assert pipeline[:-1].steps == steps[:-1]
+
+    with raises(ValueError) as e:
+        pipeline[::-1]
+
+    assert "Pipeline slicing only supports a step of 1" == str(e.value)
```
ryanSoley, line 179: as a general rule of thumb the positive (asserting correctness) and negative (asserting errors were raised) test cases are usually in separate tests. if you're concerned about duplicating code the `pipeline` object you create at the start could be a fixture like `project_client` and `fake_estimator_cls`

ryanSoley (approved):
LGTM 🎉

Issue #196, opened 2022-03-16
Include `msgpack` for Rubicon-ml pip instsall


shania-m:
**Describe the bug**
`msgpack` is a needed dependency for `intake`. Currently `intake` doesn't install `msgpack` if we use pip install for rubicon-ml. 
**Steps/Code to reproduce bug**

Run through the steps of [publish-package](https://github.com/capitalone/rubicon-ml/blob/main/.github/workflows/publish-package.yml) workflow. After you have installed the package locally with pip, run [quick-look.ipynb](https://github.com/capitalone/rubicon-ml/blob/main/notebooks/quick-look.ipynb). The notebook should fail with this error: `ModuleNotFoundError: No module named 'msgpack'` on cell 9

**Expected behavior**
After adding the dependency `msgpack` to the required section of [setup](https://github.com/capitalone/rubicon-ml/blob/main/setup.cfg), the dependency should be available in with pip install. I'd like to have this pack individual `msgpack`  added when a user installs `rubicon-ml` with `pip install rubicon-ml`

PR #199, opened 2022-03-17
update intake dependency to include `msgpack` when using `pip`

ryanSoley:
closes: #196 

---

## What
  * for some reason the regular intake install doesnt install msgpack even though its [listed in its requirements](https://github.com/intake/intake/blob/master/requirements.txt#L7)
  * installing `intake[dataframe]` fixes this
    * the issue doesn't manifest on conda install as there are no optional dependencies in conda, so the [dataframe] one is always installed


shania-m (approved):
LGTM

PR #200, opened 2022-03-23
versioneer in setup.py

ryanSoley:
## What
  * the most recent release has a nonsense version if you run `rubicon_ml --version` from the command line or `rubicon_ml.__version__` from a python interpreter
  * moves versioneer calls back into setup.py

## How to Test
  * it looked fine locally the last time so I guess we won't know until its released


shania-m (approved):
LGTM! Thanks for fixing this Ryan!

Issue #174, opened 2022-02-01
update existing `intake` catalogs


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

after #172, the experiments represented by an `intake` catalog may span multiple filesystems. the current user may not have access to all of these sources, but may want to update the catalog to add an experiment regardless. in this scenario, simply creating a new catalog may not be feasible since the user can not access all of the experiments in the catalog to be updated/replaced

**Describe the solution you'd like**

users should be able to update existing `intake` catalogs without needing to fully recreate them

**Additional context**

does intake already expose this functionality itself? let's see if there's something we can just leverage from their library before we implement something from scratch

PR #202, opened 2022-03-24
publish experiments from experiment table

ryanSoley:
## What
  * adds a button to the experiment table visualization that allows users to publish a catalog of selected experiments
    * adds a modal triggered by the button that allows input to define a catalog output path
  * updates tests to reflect changes

## How to Test
  * run the experiment table notebook and validate that the publish button writes a catalog of the selected experiments to the location you gave it

### DIFF
--- a/rubicon_ml/viz/experiments_table.py
+++ b/rubicon_ml/viz/experiments_table.py
@@ -1,7 +1,10 @@
+import os
+
 import dash_bootstrap_components as dbc
 from dash import dash_table, dcc, html
 from dash.dependencies import ALL, Input, Output, State
 
+from rubicon_ml import publish
 from rubicon_ml.viz.base import VizBase
 from rubicon_ml.viz.common.colors import light_blue, plot_background_blue
 
@@ -46,6 +49,16 @@ class ExperimentsTable(VizBase):
                     color="primary",
                     disabled=not self.is_selectable,
                     id="clear-all-button",
+                    outline=True,
+                ),
+                className="bulk-select-button-container",
+            ),
+            html.Div(
+                dbc.Button(
+                    "publish selected",
+                    color="primary",
+                    disabled=not self.is_selectable,
+                    id="publish-selected-button",
                     outline=True,
                 ),
                 className="bulk-select-button-container",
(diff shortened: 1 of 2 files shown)


rubicon_ml/viz/experiments_table.py
```diff
@@ -135,6 +148,29 @@ def layout(self):
...
+                    dbc.ModalTitle("publish selected experiments"),
+                    close_button=True,
+                ),
+                dbc.ModalBody(
+                    [
+                        dbc.Label("catalog YAML output path"),
+                        dbc.Input(
+                            id="publish-path-input",
+                            type="text",
+                            value=os.path.join(os.getcwd(), "rubicon-ml-catalog.yml"),
```
shania-m, line 163: is it possible to have the user define the name of the catalog file?

rubicon_ml/viz/experiments_table.py
```diff
@@ -314,3 +351,43 @@ def update_selected_experiment_table_rows(
...
+
+            if last_publish_selected_click > last_publish_click:
+                return True
+            elif last_publish_click > last_publish_selected_click:
+                selected_experiment_ids = [data[row].get("id") for row in selected_rows]
+                selected_experiments = [
+                    e for e in self.experiments if e.id in selected_experiment_ids
+                ]
+
+                publish(selected_experiments, publish_path)
```
shania-m, line 389: also I noticed each publish overwrites the other, could we have it appended?

rubicon_ml/viz/experiments_table.py
ryanSoley: you can change the text in the modal that pops up to change the path. is there some kind of messaging I should add to make that more obvious?

rubicon_ml/viz/experiments_table.py
ryanSoley: we have #174 for updating catalogs - once thats done we could leverage it in here

rubicon_ml/viz/experiments_table.py
shania-m: Since it was gray I assumed it was grayed out - could be just me. To be safe we could try doing something like `eneter catalog YAML output path here`. Whatever you decide works.

ryanSoley: gotcha - I updated the text and also made the text box bigger so it can fully show longer paths

shania-m: awesome thanks!

shania-m (approved):
LGTM 🎊 🎉

PR #211, opened 2022-04-08
remove `hvplot` dependency

ryanSoley:
## What
  * removes the `hvplot` dependency from the dataframe client
    * `dataframe.plot` still works, just uses plotly express now since we've moved from holoviews to dash & plotly
    * this should also fix the problem that the edgetest action is having

## How to Test
  * run the updated "visualizing-logged-dataframes.ipynb" notebook


shania-m (dismissed):
💯 🥇

shania-m (approved):
LGTM

shania-m (approved):
LGTM 💯

PR #220, opened 2022-04-12
clean up config & envs

ryanSoley:
## What
  * reconciles differences between setup.cfg and environments
  * removes old `ProjectSource` entry point that should've been removed in #181 
  * standardizes environments
  * updates pre-commit tool versions

## How to Test
  * make a new dev env and make sure the tests pass


shania-m (approved):
LGTM

Issue #191, opened 2022-03-03
Artifact Plots in Readable formats


shania-m:
**Is your enhancement request related to a problem? Please describe**
In the current implementation, when users artifact plots it is not in a human readable format. A user needs to take additional steps to view the plots. 

**Describe the solution you'd like**
Have artifact plots stored in a readable format like PNGs. This will eliminate the need to take additional steps when sharing plots to be made human readable.

ryanSoley:
this ticket needs a lot more detail but I think it'll end up being a new example in the documentation. I'm gonna move it to the docs milestone for now

we can already log PNGs as artifacts so if they're asking for something else we'll have to figure that out

shania-m:
Our customer's use case requires a collection of plots to be logged with each feature like so:   `{feature_name: png_bytes}`. They use pickle.dumps() for serialization and logging with rubicon. This requires extra effort to structure the dictionaries to make sure that the same concept  (e.g. SHAP feature importance) are logged together . Additionally, extra effort is spent on serializing and unserializing the plots for viewing. 


They are not utilizing the current logging feature for Rubicon since it requires the effort to create a file for the png, then log the file, and finally delete the file. For our customer's optimization they are using in storing the PNGSs memory strings and using the `data_bytes` argument.   If they were to log each as an individual artifact, then they could have 100+ feature importance artifacts as opposed to just one to log and then retrieve.

PR #218, opened 2022-04-11
added example for artifacting plots

shania-m:
addresses example for: #191

---

## What
  * provides Artifacting examples for plots in Rubicon

PR #222, opened 2022-04-20
Logging feature plots

shania-m:
closes: #191

---

## What
  * New example showing how to log features with their corresponding plots

## How to Test
  * run the new notebook 
  * Merge #218 first


environment.yml
```diff
@@ -33,6 +33,9 @@ dependencies:
   - pytest
   - pytest-cov
   - python-kaleido
+  - shap
+  - numba=0.55.0
```
ryanSoley, line 37: do we need the hard pin here?

ryanSoley:
sorry, its been a while since I've taken a look at this and I'm a bit confused. this says it closes #191, but it looks like #218 has already addressed #191. #191 doesn't have anything to do with features. is there an issue somewhere detailing the work necessary for this feature example? (if not, my bad - that's on me for not splitting these two issues properly)

I imagined the goal of this example was to show how to associate artifacts to features on a single experiment. for example, I have a single experiment with a number of features (`n`) I'm interested in. I do some training with a particular set of input parameters and calculate the SHAP values for each feature based on those inputs. this, to me, still seems like one experiment. that one experiment will have a plot for each feature - `n` features, `n` artifacts. then, in the future, the experimenter can compare the `n` feature plots from this experiment (we'll call "A") to the feature plots logged to an experiment "B" later

in the example here, we've got an experiment for each feature which (I believe - if the users have said otherwise please correct me) doesn't really make sense

basically, what I was imagining we'd see in this example is something like this
```python
experiment = project.log_experiment(name="associating artifacts to features")
# some code to log features and artifacts to `experiment`

for feature in experiment.features():
    artifact = experiment.get_artifact(name=feature.name)

    # some code to show logged artifact
```

environment.yml
shania-m: Let me see if I can do >=55.0 to get it working with Shap

ryanSoley:
the layout looks good now but I just created a new environment with the user-environment.yaml and the last cell in the notebook isn't displaying any charts

notebooks/logging-examples/logging-feature-plots.ipynb
```diff
@@ -0,0 +1,466 @@
...
+    "shap_values = explainer.shap_values(X)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "596f75d3",
+   "metadata": {},
+   "source": [
+    "## Plotting\n",
+    "The generated `shap_values` from the above cell can be passed to a `shap.depence_plot` to generate a dependence plot. `pl.gcf()` allows the plot generated by `shap` to be saved to a variable. Using the `matplotlib` and  `io` libraries, shap plots can be saved to a `byte` representation. Here, a feature and its plot are both logged with`rubicon_ml.Features` and `rubuicon_ml.Artifact` respectively. These `features` and `artifacts` are logged to the same `rubicon_ml.Experiment` which are being kept track of with an experiment tag.  "
```
ryanSoley, line 94: "which are being kept track of with an experiment tag" should be removed from the end of that last sentence

notebooks/logging-examples/logging-feature-plots.ipynb
```diff
@@ -0,0 +1,466 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8ab77a88",
+   "metadata": {},
+   "source": [
+    "## Retrieving your logged plot and features programmatically\n",
+    "\n",
+    "Finally, using the tag for each experiment, we can retrieve a feature and its associated plot. For this example we'll inspect all the `features` logged to the experiment and its associated plots using ` exp.artifact(name=feature.name)`. With `IO` and `PIL`, after retrieving the PNG byte representation of the plot, the plot can be rendered as a PNG image. "
```
ryanSoley, line 292: we're not using the tag to retrieve experiments anymore

ryanSoley (commented):
its woking now 🎉 

just a couple small things in the text and itll be good to go

ryanSoley (approved):

PR #228, opened 2022-04-26
add python-kaleido to binder for edge test

shania-m:
---

## What
  * addresses edge test failures due to python-kaleido dependency 

## How to Test
  *


srilatharanganathan (dismissed):
LGTM

ryanSoley:
while we do need the dependency in binder, thats not whats causing this to fail. python-kaleido is in edgetests section setup.cfg, so it should be getting installed
> https://github.com/capitalone/rubicon-ml/blob/main/setup.cfg#L102

@fdosani any thoughts?

fdosani:
> while we do need the dependency in binder, thats not whats causing this to fail. python-kaleido is in edgetests section in setup.cfg, so it should be getting installed
> 
> > https://github.com/capitalone/rubicon-ml/blob/main/setup.cfg#L102
> 
> @fdosani any thoughts?

Seeing if I can recreate this locally.

PR #230, opened 2022-05-02
Fixing PIL/kaleido issues for edgetest

fdosani:
closes: #228

---

## What
  * Fixes the dependencies for unit tests to run and cause edgetest to pass.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -91,6 +91,9 @@ xfail_strict = True
 
 [edgetest.envs.core]
 python_version = 3.9
+deps =
+    	kaleido
+	Pillow
 conda_install = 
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -91,6 +91,9 @@ xfail_strict = True
 
 [edgetest.envs.core]
 python_version = 3.9
+deps =
+    	kaleido
```
shania-m, line 95: its not python-kaleido?

fdosani: It is different between `conda` and `pip`.

shania-m (approved):
LGTM

PR #233, opened 2022-05-05
added feature plot dependencies to setup.cfg

shania-m:
## What
  * Added dependencies for new notebook for feature plots #222 to setup.cfg for edge tests

## How to Test
  * run `edgetest -c setup.cfg --export` to ensure edge tests pass

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -95,8 +95,8 @@ deps =
 	kaleido
 	Pillow
 	shap
-    numba>=0.55.0
-    matplotlib
+	numba>=0.55.0
+	matplotlib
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -94,6 +94,9 @@ python_version = 3.9
 deps = 
 	kaleido
 	Pillow
+	shap
+	numba>=0.55.0
```
ryanSoley, line 98: edgetest will always test the latest so we dont need any pins in here

ryanSoley (approved):
thanks!!

Issue #190, opened 2022-03-03
Identify/Remove Failed Experiments


shania-m:
**Is your feature request related to a problem? Please describe**

Rubicon does not support the removal of experiments. When a program fails, an empty experiment is created.

**Describe the solution you'd like**
Users should be able to able to remove/distinguish failed experiments.
We should see if we can accomplish this by implementing tag for experiments with a Success/Fail status.

PR #235, opened 2022-05-05
add identifying failed experiments example

shania-m:
closes: #190 

---

## What
  * added example for identifying failed experiments

## How to Tests
  * all notebook tests pass


notebooks/logging-examples/logging-experiment-failures.ipynb
```diff
@@ -0,0 +1,201 @@
...
+    "    def __init__(self, n_neighbors=2):\n",
+    "        super().__init__()\n",
+    "        \n",
+    "        self.n_neighbors = n_neighbors\n",
+    "        \n",
+    "        self.knn = KNeighborsClassifier(n_neighbors=n_neighbors)\n",
+    "    def fit(self, X, y):\n",
+    "        self.knn.fit(X, y)\n",
+    "        try:\n",
+    "            raise RuntimeError(\"Bad Estimator\")\n",
```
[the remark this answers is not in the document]
ryanSoley: per our slack convo: like you pointed out, the pipeline will not create an experiment if an estimator raises an error. I think this example can still be valuable in terms of "identifying failed experiments" tho. if we define the fit as just:
```
def fit(self, X, y):
    self.knn.fit(X, y)

    if SOME_RANDOM_CONDITION:
        self.result_ = SOME_OUTPUT
```
and then say that a "success" is when `self.result_` is populated and a "failure" is when its not, we can still do the last example in the previous comment. I think from there people can at least reason how to determine their own success or failure conditions and adapt the example to whatever use case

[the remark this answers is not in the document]
shania-m: would `self.result_` be logged? since its not a parameter to the fit, it just an attribute to the estimator itself?

notebooks/logging-examples/logging-experiment-failures.ipynb
```diff
@@ -0,0 +1,157 @@
...
+    "## Distinguishing Failed Experiments with Rubicon \n",
+    "Modeling runs can fail for a number of reasons. When logging with rubicon_ml, a failed run may result in an empty or incomplete experiment. In this example, we'll walk through how to handle such experiments."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d4f58990",
+   "metadata": {},
+   "source": [
+    "First lets simulate the problem. To do this we'll create an estimator that will fail on it's `fit()` 30% of the time.  We will consider any experiment that has a `none` value for `self.state`."
```
ryanSoley, line 17: last sentence isn't quite finished

_We will consider any pipeline that has a learned attribute `self.state_` to have "succeeded," and any that does not to have "failed."_

the experiments' success/failure depends on the attributes of the pipeline, not the experiment itself. the experiment is just tracking that info

notebooks/logging-examples/logging-experiment-failures.ipynb
```diff
@@ -0,0 +1,157 @@
...
+    "    def __init__(self, state=None):\n",
+    "        super().__init__()\n",
+    "        self.state=None\n",
+    "        self.knn = KNeighborsClassifier(n_neighbors=2)\n",
+    "    def fit(self, X, y):\n",
+    "        self.knn.fit(X, y)\n",
+    "        output=random.random()\n",
+    "        if output<.3:\n",
+    "            output=None\n",
+    "        self.state=output\n",
```
ryanSoley, line 40: `state` is a learned attribute, so it should be `self.state_` per sklearn standards

also we should just say 
```
if output > .3
    self.state_ = output
````

this way `self.state_` is the random value if its greater than 0.3 and thus a "success," or it just doesn't exist if the random value is less than 0.3 and thus a "failure"

File: notebooks/logging-examples/logging-experiment-failures.ipynb [hunk 4]
```diff
@@ -0,0 +1,157 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from sklearn.base import BaseEstimator\n",
+    "import random\n",
+    "\n",
+    "class BadEstimator(BaseEstimator):\n",
+    "    def __init__(self, state=None):\n",
+    "        super().__init__()\n",
+    "        self.state=None\n",
```
ryanSoley, line 33: we don't need to set this to `None` in the constructor. learned attributes like `self.state_` are either available as attributes after a call to `fit` or don't exist as attributes on the object at all

for example, pre-`fit` `bad_estimator_instance.state_` would error

notebooks/logging-examples/logging-experiment-failures.ipynb
```diff
@@ -0,0 +1,157 @@
...
+    "\n",
+    "rubicon = Rubicon(\n",
+    "    persistence=\"memory\",\n",
+    ")\n",
+    "project = rubicon.get_or_create_project(name=\"Failed Experiments\")\n",
+    "\n",
+    "X = [[1], [1], [1], [1]]\n",
+    "y = [1, 1, 1, 1]\n",
+    "pipe=make_pipeline(project, SimpleImputer(strategy=\"mean\"),BadEstimator())\n",
+    "for _ in range(20):\n",
```
ryanSoley, line 78: can we move this for loop into the next cell and combine it with that loop? something like this in the following cell:

```
for _ in range(20):
    pipe.fit(X, y)

    if not hasattr(pipe, "state_"):
        pipe.experiment.add_tags(["failed"])

File: notebooks/logging-examples/logging-experiment-failures.ipynb [hunk 6]
```diff
@@ -0,0 +1,157 @@
...
+      "Experiment(project_name='Failed Experiment', id='a8d2bbef-ab06-450b-b73c-acdfaaaf6e9e', name='RubiconPipeline experiment', description=None, model_name=None, branch_name=None, commit_hash=None, training_metadata=None, tags=['failed'], created_at=datetime.datetime(2022, 5, 9, 18, 54, 12, 526703))\n",
+      "Experiment(project_name='Failed Experiment', id='301c11a4-cfc9-4f5c-900e-35f55d38c539', name='RubiconPipeline experiment', description=None, model_name=None, branch_name=None, commit_hash=None, training_metadata=None, tags=['failed'], created_at=datetime.datetime(2022, 5, 9, 18, 54, 12, 539486))\n",
+      "Experiment(project_name='Failed Experiment', id='a228c707-39b8-4bf0-9813-64cf51e7036e', name='RubiconPipeline experiment', description=None, model_name=None, branch_name=None, commit_hash=None, training_metadata=None, tags=['failed'], created_at=datetime.datetime(2022, 5, 9, 18, 54, 12, 541330))\n"
+     ]
+    }
+   ],
+   "source": [
```
ryanSoley, line 130: can we expand this cell a little to show the percentage of failed and successful experiments - basically want to show that its about 70% success/30% fail

notebooks/logging-examples/logging-experiment-failures.ipynb
shania-m: Isn't state an attribute of the estimator itself, not the pipeline? Also, when I don't have it in the constructor I don't see it logged

notebooks/logging-examples/logging-experiment-failures.ipynb
shania-m: Since we're looping and fitting multiple times, I'm keeping resetting it to `none` otherwise it would keep the `state
_` value of the last success

notebooks/logging-examples/logging-experiment-failures.ipynb
ryanSoley: yeah my bad, you'll need to access it from the estimator. so something like `pipeline.stages[0].state_` or `pipeline["stage_name"].state_`

ryanSoley (approved):
awesome! thanks for all the changes

Issue #252, opened 2022-07-21
new quick look notebooks failing in `edgetest`


ryanSoley:
**Describe the bug**

#249 introduced a new quick look notebook, which does a `pip install` in its first cell to download a dataset. this `pip install` works fine when running the tests via our test workflow, but seems to not work within the edgetest workflow

**Steps/Code to reproduce bug**

here's the most recent failed run
> https://github.com/capitalone/rubicon-ml/runs/7447008762?check_suite_focus=true

**Expected behavior**

tests to behave the same (both pass) via the test and edgetest workflows

PR #253, opened 2022-07-21
use `conda` in `edgetest`

ryanSoley:
closes: #252 
---

## What
  * updates edgetest to use the `update_with_conda` flag recently introduced in [edgetest-conda](https://github.com/capitalone/edgetest-conda/)
    * this should resolve the missing dependency noted in #252 
  * adds the `workflow_dispatch` trigger to the edgetest workflow so future changes can be tested off-schedule

## How to Test
 I think we'll need to merge before we can test, unless the `workflow_dispatch` trigger shows up just via the PR
  * yeah this looks to be the case - I dont see the `workflow_dispatch` dropdown on the edgetest action yet

### DIFF
--- a/.github/workflows/edgetest.yml
+++ b/.github/workflows/edgetest.yml
@@ -5,6 +5,8 @@ name: Run edgetest
 on:
   schedule:
     - cron: '0 10 * * 1-5'
+  workflow_dispatch:
+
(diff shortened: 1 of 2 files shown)


andreafehrman (approved):
LGTM!

PR #257, opened 2022-07-28
edgetest update

ryanSoley:
## What
  * applies the good version updates from https://github.com/capitalone/rubicon-ml/pull/256
    * that PR was opened a few days ago and eventually closed itself because `prefect` v2.0.0 caused tests to start failing
    * only applies the updates before `prefect` v2.0.0
  * moves `prefect` from the `update` section of the `edgetest` config to the `conda_install` section
    * this will no longer attempt to update `prefect`, effectively locking it in at less than v2.0.0
    * we can revisit this later if necessary
  * removes the `ref: main` tag from the checkout action so the manual trigger can select non-main branches
  * applies the version updates from https://github.com/capitalone/rubicon-ml/pull/258
    * these were generated after fixing the manual trigger and getting a successful test run

## How to Test
  * original tests on https://github.com/capitalone/rubicon-ml/pull/256 already passed
  * RESOLVED ~manually triggering the workflow seems to always run off of main regardless of what branch is selected, so we'll have to wait until the action is merged to test it~
    * ~I've reached out to the `edgetest` team to see how to properly run against non-main branches~
  * [the action successfully ran against this new branch](https://github.com/capitalone/rubicon-ml/runs/7561040207?check_suite_focus=true)

### DIFF
--- a/.github/workflows/edgetest.yml
+++ b/.github/workflows/edgetest.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          ref: main
+          fetch-depth: 0
       - id: run-edgetest
         uses: fdosani/run-edgetest-action@v1.1
         with:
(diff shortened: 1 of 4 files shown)


andreafehrman (approved):
LGTM

PR #261, opened 2022-08-26
fix whitesource failures

ryanSoley:
## What
  * adds `numpy` to setup.cfg and edgetest config to hopefully stop whitesource from [failing](https://github.com/capitalone/rubicon-ml/runs/7952837318)
    * ~`numpy` 1.23 is not yet compatible with `numba`, so we're pinned to 1.22 only here rn~
      * ~looks like it should be soon (https://github.com/numba/numba/issues/8263)~
      * addressed by `numba`
  * also adds `scikit-learn` since it was the only other top-level dependency not present other than `numpy`

## How to Test
  * I confirmed that the `numpy` and `sklearn` lower bounds pass tests with the python version lower bound
  * [edgetest is passing](https://github.com/capitalone/rubicon-ml/runs/8255712989?check_suite_focus=true)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
 	dask[dataframe]<=2022.8.1,>=2.12.0
 	fsspec<=2022.7.1,>=2021.4.0
 	intake[dataframe]<=0.6.5,>=0.5.2
-	numpy<=1.22.4,>=1.22.0
+	numpy<=1.23.2,>=1.22.0
 	pandas<=1.4.3,>=1.0.0
 	pyarrow<=9.0.0,>=0.18.0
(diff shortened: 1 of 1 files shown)


ryanSoley:
[running the edgetest action](https://github.com/capitalone/rubicon-ml/actions/runs/2935072786) to test if the changes work have manifested an issue with the most recent version of `intake` (0.6.6). that will need to be fixed before continuing here

ryanSoley:
> [running the edgetest action](https://github.com/capitalone/rubicon-ml/actions/runs/2935072786) to test if the changes work has manifested an issue with the most recent version of `intake` (0.6.6). that will need to be fixed before continuing here

resolved in #262

shania-m (approved):
LGTM

Issue #244, opened 2022-06-02
Add Tags to Artifact Class


shania-m:
**Is your enhancement request related to a problem? Please describe**
A clear and concise description of what the problem is. Ex. I wish I could use Rubicon to do [...].
An artifact is a catch-all for any other type of data that can be logged to a file. I'd like users to be easily able to distinguish artifacts since they can be vague in nature. 

**Describe the solution you'd like**
Users should be able to tag experiments for retrieval/logging. For example if a user logs multiple plots each plot can have a tag in the artifact. Like this:
```
bar_plot = px.bar()
bar_plot_bytes=bar_plot.to_image(format="png")
exp.log_artifact(data_bytes=bar_plot_bytes,tags=["bar plot"])
line_plot = px.line()
line_plot_bytes=line_plot.to_image(format="png")
exp.log_artifact(data_bytes=bar_plot_bytes,tags=["line plot"])

```
**Describe alternatives you've considered**

The `name` and `description` attributes can be used as well - however since experiments and dataframes both have tags we should introduce tags to Artifacts. 
**Additional context**
Add any other context, code examples, or references to existing implementations about the feature request here.

> Tags were implemented as mixin on the client and domain
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/mixin.py#L409
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/domain/mixin.py#L4
> Extend the artifact classes in the client and domain with those classes like in the dataframe here
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/dataframe.py#L7
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/domain/dataframe.py#L12

PR #268, opened 2022-09-13
add tagging to artifacts

ryanSoley:
closes: #244 

---

## What
  * add tagging to artifacts
    * applied the `TagMixin` to the artifact domain object
    * updated client and repo layer to reflect `TagMixin` addition
      * updated `_get_tag_metadata_root` to handle all rubicon objects
  * updated tests

## How to Test
  * run through the quick look examples and validate they work as expected
  * try logging tags to artifacts
 
```python

rubicon = Rubicon(persistence="filesystem", root_dir=".")

# test tagging project artifacts
project = rubicon.get_or_create_project(name="test artifacts")
project.log_artifact(data_bytes=b"bytes", name="name", tags=["tags_a"])

# test tagging experiment artifacts
experiment = project.log_experiment()
experiment.log_artifact(data_bytes=b"bytes", name="name", tags=["tags_b"])
```

### DIFF
--- a/rubicon_ml/client/asynchronous/mixin.py
+++ b/rubicon_ml/client/asynchronous/mixin.py
@@ -65,7 +65,7 @@ class ArtifactMixin:
 
         artifact = domain.Artifact(name=name, description=description, parent_id=self._domain.id)
 
-        project_name, experiment_id = self._get_parent_identifiers()
+        project_name, experiment_id = self._get_identifiers()
         await self.repository.create_artifact(
             artifact, data_bytes, project_name, experiment_id=experiment_id
         )
@@ -129,7 +129,7 @@ class ArtifactMixin:
         list of rubicon.client.Artifact
             The artifacts previously logged to this client object.
         """
-        project_name, experiment_id = self._get_parent_identifiers()
+        project_name, experiment_id = self._get_identifiers()
 
         self._artifacts = [
             client.Artifact(a, self)
@@ -150,7 +150,7 @@ class ArtifactMixin:
         ids : list of str
             The ids of the artifacts to delete.
         """
-        project_name, experiment_id = self._get_parent_identifiers()
+        project_name, experiment_id = self._get_identifiers()
 
         await asyncio.gather(
             *[
@@ -186,7 +186,7 @@ class DataframeMixin:
         """
         dataframe = domain.Dataframe(parent_id=self._domain.id, description=description, tags=tags)
 
-        project_name, experiment_id = self._get_parent_identifiers()
+        project_name, experiment_id = self._get_identifiers()
         await self.repository.create_dataframe(
             dataframe, df, project_name, experiment_id=experiment_id
         )
@@ -225,7 +225,7 @@ class DataframeMixin:
         list of rubicon.client.Dataframe
             The dataframes previously logged to this client object.
         """
-        project_name, experiment_id = self._get_parent_identifiers()
+        project_name, experiment_id = self._get_identifiers()
         dataframes = [
             client.Dataframe(d, self)
             for d in await self.repository.get_dataframes_metadata(
@@ -247,7 +247,7 @@ class DataframeMixin:
         ids : list of str
             The ids of the dataframes to delete.
         """
-        project_name, experiment_id = self._get_parent_identifiers()
+        project_name, experiment_id = self._get_identifiers()
 
         await asyncio.gather(
             *[
--- a/rubicon_ml/client/asynchronous/mixin.py
+++ b/rubicon_ml/client/asynchronous/mixin.py
@@ -3,11 +3,10 @@ from datetime import datetime
 
 from rubicon_ml import domain
 from rubicon_ml.client import asynchronous as client
-from rubicon_ml.client.mixin import MultiParentMixin
 from rubicon_ml.client.utils.tags import has_tag_requirements
 
 
-class ArtifactMixin(MultiParentMixin):
+class ArtifactMixin:
     """Adds artifact support to an asynchronous client object."""
 
     async def log_artifact(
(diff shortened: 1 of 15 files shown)


rubicon_ml/client/mixin.py
```diff
@@ -410,15 +424,18 @@ class TagMixin:
...
-        dataframe_id = None
+        entity_id = None
 
-        if isinstance(self, client.Dataframe):
-            project_name, experiment_id = self.parent._get_parent_identifiers()
-            dataframe_id = self.id
-        else:
+        if isinstance(self, client.Project):
             project_name, experiment_id = self._get_parent_identifiers()
+        elif isinstance(self, client.Experiment):
```
shania-m, line 431: whats this case for?

shania-m: do we need to grab entity id for experiment as well?

shania-m: add a comment explaining what is being accomplished

rubicon_ml/repository/base.py
```diff
@@ -913,19 +913,30 @@ def get_parameters(self, project_name, experiment_id):
 
     # ---------- Tags ----------
 
-    def _get_tag_metadata_root(self, project_name, experiment_id=None, dataframe_id=None):
-        if dataframe_id is not None:
-            dataframe_metadata_root = self._get_dataframe_metadata_root(project_name, experiment_id)
+    def _get_tag_metadata_root(
+        self, project_name, experiment_id=None, entity_id=None, entity_type=None
+    ):
+        get_metadata_root_lookup = {
```
shania-m, line 919: nice!!

rubicon_ml/repository/base.py
```diff
@@ -964,7 +977,9 @@ def remove_tags(self, project_name, tags, experiment_id=None, dataframe_id=None)
             The ID of the dataframe to delete the tags
             `tags` from.
         """
```
shania-m, line 979: add doc strings for `entitity_id` and `entity_type`

rubicon_ml/repository/base.py
```diff
@@ -913,19 +913,30 @@ def get_parameters(self, project_name, experiment_id):
...
             return f"{experiment_metadata_root}/{experiment_id}"
         else:
-            raise ValueError("`experiment_id` and `dataframe_id` can not both be `None`.")
+            entity_metadata_root = get_metadata_root(project_name, experiment_id)
+
+            return f"{entity_metadata_root}/{entity_id}"
 
-    def add_tags(self, project_name, tags, experiment_id=None, dataframe_id=None):
+    def add_tags(self, project_name, tags, experiment_id=None, entity_id=None, entity_type=None):
         """Persist tags to the configured filesystem.
```
shania-m, line 940: add docs strings for `entity_id` `entity_type`

rubicon_ml/repository/base.py
```diff
@@ -983,7 +998,7 @@ def _sort_tag_paths(self, tag_paths):
 
         return tag_paths_with_timestamps
 
-    def get_tags(self, project_name, experiment_id=None, dataframe_id=None):
+    def get_tags(self, project_name, experiment_id=None, entity_id=None, entity_type=None):
```
shania-m, line 1001: same

ryanSoley:
@shania-m you were right - that identifier stuff was confusing. the `MultiParentMixin` didn't make much sense, so I decided to remove it and put the function on the project and experiment clients instead. since the function was checking the instance of the class and doing something different for each one, I think it makes sense to be on the classes themselves instead of tucked in the mixin

I also changed the name of that function from `get_parent_identifiers` to just `get_identifiers`, thats why there's so many more files in the diff now

rubicon_ml/client/mixin.py
ryanSoley: addressed in comment below

rubicon_ml/client/asynchronous/mixin.py
```diff
@@ -3,11 +3,10 @@
 
 from rubicon_ml import domain
 from rubicon_ml.client import asynchronous as client
-from rubicon_ml.client.mixin import MultiParentMixin
 from rubicon_ml.client.utils.tags import has_tag_requirements
 
 
-class ArtifactMixin(MultiParentMixin):
```
ryanSoley, line 10: this was never needed - removing

rubicon_ml/client/mixin.py
```diff
@@ -10,27 +10,7 @@
...
-            project_name = self.name
-        else:
-            project_name = self.project.name
-            experiment_id = self.id
-
-        return project_name, experiment_id
-
-
-class ArtifactMixin(MultiParentMixin):
+class ArtifactMixin:
```
shania-m, line 13: this looks SO much better

rubicon_ml/client/mixin.py
```diff
@@ -410,15 +404,16 @@ class TagMixin:
...
-        dataframe_id = None
+        project_name, experiment_id = self._parent._get_identifiers()
+        entity_id = None
 
-        if isinstance(self, client.Dataframe):
-            project_name, experiment_id = self.parent._get_parent_identifiers()
-            dataframe_id = self.id
+        # experiments are not required to return an entity ID - they are the entity
+        if isinstance(self, client.Experiment):
+            experiment_id = self.id
```
shania-m, line 412: what is experiment_id if this line doesn't execute?

ryanSoley: its the second value returned from `self._parent._get_identifiers()` on line 407. in this case, `None` specifically

it needs this experiment case because `get_identifiers` on a project (a project is always the experiment's parent) will return `None` for `experiment_id` since a project of course doesn't have an experiment ID

but, this function still needs the experiment's ID, so it just gets it from `self.id`. any other class that we're tagging that requires an experiment ID will have it returned from `self._parent._get_identifiers()` before we get into the if-else as those will all have experiments for parents, not projects

shania-m: sounds good!

shania-m (approved):
LGTM

PR #276, opened 2022-09-22
name cleanup

ryanSoley:
## What
  * cleaning up/standardizing some naming after the last few tagging stories
    * `filter_entity` -> `filter_children`
    * this ones pretty trivial, but `entity_id` -> `entity_identifier` just to make it clear that the "identifier" can be an id or a name

## How to Test
  * `python -m pytest`


Tejass9922 (commented):
Looks good

joe-wolfe21 (approved):

Issue #272, opened 2022-09-19
add tagging to parameter class


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

some experiments have many parameters logged to them - retrieving them via tags, like experiments, dataframes, and artifacts, would allow users to pull associated groups of parameters back off an experiment all at once

**Describe the solution you'd like**

parameters should be able to be logged with tags
```python
parameter = experiment.log_parameter(name="k_neighbors", value=2, tags=["small"])
parameter.add_tags(tags=["more", "tags"])
```

parameters should be able to be retrieved by tags
```python
parameters = experiment.parameters(tags=["small"])
```

**Describe alternatives you've considered**

parameters can be retrieved by name, but retrieving large groups by their individual names can be cumbersome

**Additional context**

this should be able to be implemented in the same style as #244, leveraging the mixins below
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/domain/mixin.py#L4
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/mixin.py#L403

PR #280, opened 2022-09-27
272 parameter tagging

armaan-dhillon:
closes: #272

---

## What
  * Adds tagging functionality to parameters in Rubicon experiments
       *  Extends parameter domain with `TagMixin` class
       *  Updates parameter functions in `client/experiment.py` for tagging functionality
       *  Adds `parents` attribute to parameter client

## How to Test
  * Run unit tests with pytest
  * Open quick-look jupyter notebook to test functionality

> ```
> from rubicon_ml import Rubicon
> 
> rubicon = Rubicon(
>     persistence="filesystem",
>     root_dir="./rubicon-root",
>     auto_git_enabled=True,
> )
> project = rubicon.get_or_create_project(name="classifying penguins")
> experiment = project.log_experiment()
> 
> param_a = experiment.log_parameter(name="param_a")
> param_b = experiment.log_parameter(name="param_b")
> param_c = experiment.log_parameter(name="param_c")
> 
> param_a.add_tags(["tag1", "tag2"])
> param_b.add_tags(["tag1", "tag3"])
> param_c.add_tags(["tag2"])
> 
> param_a.tags
> param_b.tags
> 
> param_b.remove_tags(["tag3"])
> param_b.tags
> 
> parameters_list = experiment.parameters(tags=["tag1"])
> parameters_list
> ```

### DIFF
--- a/rubicon_ml/client/experiment.py
+++ b/rubicon_ml/client/experiment.py
@@ -217,7 +217,7 @@ class Experiment(Base, ArtifactMixin, DataframeMixin, TagMixin):
             The parameter's description. Use to provide
             additional context.
         tags : list of str, optional
-            Values to tag the experiment with. Use tags to organize and
+            Values to tag the parameter with. Use tags to organize and
             filter your parameters.
 
         Returns
--- a/rubicon_ml/client/experiment.py
+++ b/rubicon_ml/client/experiment.py
@@ -202,7 +202,7 @@ class Experiment(Base, ArtifactMixin, DataframeMixin, TagMixin):
 
         return feature
 
-    def log_parameter(self, name, value=None, description=None):
+    def log_parameter(self, name, value=None, description=None, tags=[]):
         """Create a parameter under the experiment.
 
         Parameters
@@ -216,29 +216,44 @@ class Experiment(Base, ArtifactMixin, DataframeMixin, TagMixin):
         description : str, optional
             The parameter's description. Use to provide
             additional context.
+        tags : list of str, optional
+            Values to tag the experiment with. Use tags to organize and
+            filter your parameters.
 
         Returns
         -------
         rubicon.client.Parameter
             The created parameter.
         """
-        parameter = domain.Parameter(name, value=value, description=description)
+        parameter = domain.Parameter(name, value=value, description=description, tags=tags)
         self.repository.create_parameter(parameter, self.project.name, self.id)
 
-        return Parameter(parameter, self._config)
+        return Parameter(parameter, self)
 
-    def parameters(self):
+    def parameters(self, name=None, tags=[], qtype="or"):
         """Get the parameters logged to this experiment.
+
+        Parameters
+        ----------
+        name : str, optional
+            The name value to filter results on.
+        tags : list of str, optional
+            The tag values to filter results on.
+        qtype : str, optional
+            The query type to filter results on. Can be 'or' or
+            'and'. Defaults to 'or'.
 
         Returns
         -------
         list of rubicon.client.Parameter
             The parameters previously logged to this experiment.
         """
-        self._parameters = [
-            Parameter(p, self._config)
-            for p in self.repository.get_parameters(self.project.name, self.id)
+
+        parameters = [
+            Parameter(p, self) for p in self.repository.get_parameters(self.project.name, self.id)
         ]
+
+        self._parameters = filter_children(parameters, tags, qtype, name)
 
         return self._parameters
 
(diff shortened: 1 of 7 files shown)


Tejass9922:
pytests ran with no failures, 2 things for the notebooks testing: 
- `for feature_name in feature_names:
       experiment.log_feature(name=feature_name)` 
     this line of code doesn't explicitly test parameter tagging and also requires all previous cells in the logging experiments 
     notebook to be run before penguins data set to run, consider potentially taking it out

- `parameters_list = experiment.parameters(tags=["tag1"])` this line gives multiple warnings about how the metadata json is not found for your parameter variables

rubicon_ml/client/experiment.py
```diff
@@ -216,30 +216,45 @@ def log_parameter(self, name, value=None, description=None):
         description : str, optional
             The parameter's description. Use to provide
             additional context.
+        tags : list of str, optional
+            Values to tag the experiment with. Use tags to organize and
```
ryanSoley, line 220:
```suggestion
            Values to tag the parameter with. Use tags to organize and
```

rubicon_ml/client/parameter.py
```diff
@@ -18,12 +18,14 @@ class Parameter(Base):
     ----------
     domain : rubicon.domain.Parameter
         The parameter domain model.
-    config : rubicon.client.Config
-        The config, which specifies the underlying repository.
+    parent : rubicon.client.Experiment
+        The experiment that the metric is logged to.
```
ryanSoley, line 22:
```suggestion
        The experiment that the parameter is logged to.
```

rubicon_ml/client/parameter.py
```diff
@@ -18,12 +18,14 @@ class Parameter(Base):
...
     """
 
-    def __init__(self, domain, config=None):
-        super().__init__(domain, config)
+    def __init__(self, domain, parent):
+        super().__init__(domain, parent._config)
+        self.data = None
```
ryanSoley, line 27: this doesn't need `data`, that's just for dataframes and artifacts

ryanSoley:
> * `parameters_list = experiment.parameters(tags=["tag1"])` this line gives multiple warnings about how the metadata json is not found for your parameter variables

good catch - this is definitely a bug with metrics and features too that we didn't catch in their reviews. that warning was introduced so that when a non-rubicon-created directory ends up in a user's filesystem (like a Mac's `.DS_Store` or something) it would just be ignored instead of throwing a `metadata.json` not found error

if you take a look at the parameters directory under the experiment created by the above example you'll see six directories instead of three:
```
|- features
  |- param_a
  |- param_b
  |- param_c
  |- parama
  |- paramb
  |_ paramc
```

the parameters themselves are in the ones without the underscore. the ones with the underscore only have the updated tag files in them, which is why the three of them are raising the warning.

the parameters don't have underscores because we [`slugify`](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/repository/utils/slugify.py) their names (and metrics and features too) in the path we write to. here's `_get_parameter_metadata_path` - see its `slugify`-ing on line 825
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/repository/base.py#L825

we aren't `slugify`-ing the name in `_get_tag_metadata_root`. we should be right here in the else case
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/repository/base.py#L937

there is a caveat tho - we only want to `slugify` names, not IDs, and that `entity_identifier` can be a name or an ID. so we'll only want to `slugify` it if its a metric, feature, or parameter - not an artifact or experiment

we must have only tested metrics and features with one-word tags with no underscores or dashes. also this only happens when you add/remove tags to/from things with dashes in their names. that's why the tests are still passing even though they have underscores in them on this branch - its fine when the tags are created when the feature is logged

rubicon_ml/repository/base.py
```diff
@@ -937,6 +938,9 @@ def _get_tag_metadata_root(
         else:
             entity_metadata_root = get_metadata_root(project_name, experiment_id)
 
+            # We want to slugify the names of Metrics, Features, and Parameters- not Artifacts, Dataframes, or Experiments
+            if entity_type == "Metric" or entity_type == "Feature" or entity_type == "Parameter":
```
ryanSoley, line 942: nice! can we do `if entity_type in ["Metric", "Feature", "Parameter"]:` instead? its a bit cleaner imo once we have more than two and/or clauses

PR #288, opened 2022-10-12
remove hard `dask` requirement

ryanSoley:
## What
  * rubicon-ml currently has a hard dask requirement defined in its setup.cfg
    * this is so users can seamlessly log both `dask` and `pandas` dataframes to rubicon-ml
  * the `dask` library is only leveraged directly by rubicon-ml when reading dataframes specified as the "dask" type
    * the "dask" type is never the default, "pandas" is - and `pandas` is still a hard requirement
    * a new error is thrown if users attempt to read a dataframe with a type of "dask" without the `dask` library installed
  * logging `dask` dataframes requires users to already have one to log, so they must have installed `dask` themselves
  * `dask` is still included in the environment as a testing dependency so the tests can still leverage it
    * tests have been updated to ensure new methods work with `dask` and `pandas` dataframes

## How to Test
  * create a new dev environment, then run the tests: `python -m pytest`
  * create a new dev environment, then run the concurrent logging notebook - this one leverages the `to_df` function

### DIFF
--- a/docs/docs-environment.yml
+++ b/docs/docs-environment.yml
@@ -19,7 +19,6 @@ dependencies:
   # so local pip install doesn't
   - dash<=2.6.2,>=2.0.0
   - dash-bootstrap-components<=1.2.1,>=1.0.0
-  - dask[dataframe]<=2022.9.2,>=2.12.0
   - fsspec<=2022.8.2,>=2021.4.0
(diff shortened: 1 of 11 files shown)


Tejass9922 (approved):
Looks good! pytest passes and concurrent logging notebook works as well!

PR #308, opened 2022-11-28
update existing intake catalogs

Tejass9922:
closes: #174 

---

## What
  *  Added functionality to update existing `intake` catalogs. 
  * Instead of calling  the `open_catalog`, `discover` and `read` functions on the yaml file in order to read and then publish the new experiments to a catalog, we now only need to call the `publish` function with an additional parameter: `catalog_filepath` 
  * `publish` function upon seeing this new argument will call helper function `update_catalog` which handles stitching new experiments to original yaml file 

## How to Test
  * run pytest
  * view and run cell 4 in `quick-look/sharing_experiments` notebook


rubicon_ml/intake_rubicon/publish.py
```diff
@@ -18,13 +19,24 @@ def publish(
         and key to log the generated YAML file to. S3 buckets
         must be prepended with 's3://'. Defaults to None,
         which disables writing the generated YAML.
+    base_catalog_filepath : str, optional
+        Similar to output_filepath except this argument is used as a
+        base fileto update exisiting intake catalog. Defaults to None,
```
ryanSoley, line 24:
```suggestion
        Similar to `output_filepath` except this argument is used as a
        base file to update exisiting intake catalog. Defaults to None,
```

just nitpicking - use backticks around argument names in the docstrings

rubicon_ml/intake_rubicon/publish.py
```diff
@@ -18,13 +19,24 @@ def publish(
...
         The YAML string representation of the `intake` catalog
         containing the experiments `experiments`.
     """
+
+    if base_catalog_filepath is not None:
+        return update_catalog(
+            base_catalog_filepath=base_catalog_filepath,
+            new_experiments=experiments,
+            output_filepath=output_filepath,
+        )
```
ryanSoley, line 39: more nitpicking - can you add a newline after this. I'm surprised `black` didn't complain about that

rubicon_ml/intake_rubicon/publish.py
```diff
@@ -47,3 +59,60 @@ def publish(
             f.write(catalog_yaml)
 
     return catalog_yaml
+
+
+"""Helper function to update exisiting intake catalog.
```
ryanSoley, line 64: this should be inside `update_catalog` on line 91

rubicon_ml/intake_rubicon/publish.py
```diff
@@ -47,3 +59,60 @@ def publish(
...
+        experiment_catalog_name = f"experiment_{experiment.id.replace('-', '_')}"
+        updated_catalog[experiment_catalog_name] = appended_experiment_catalog
+
+    with fsspec.open(base_catalog_filepath, "r") as yamlfile:
+        curr_catalog = yaml.safe_load(yamlfile)
+
+        curr_catalog["sources"].update(updated_catalog)
+
+    resulting_filepath = base_catalog_filepath if not output_filepath else output_filepath
+    if curr_catalog:
```
ryanSoley, line 112: is this `if` statement necessary? `curr_catalog` will always not be None, right?

ryanSoley (commented):
there are a few chunks of duplicated code in the `publish` and `update_catalog` now - like the loop over the list of experiments and the final catalog writing in both. can we move those into functions?

Tejass9922:
> there are a few chunks of duplicated code in the `publish` and `update_catalog` now - like the loop over the list of experiments and the final catalog writing in both. can we move those into functions?

good point! we can modularize the reading of experiments and "catalogging" process . For the final catalog writing and all other catalog writing actually, there's 3 types of writes / reads and all of them are different even though that initial `with open` is the same so I'm not sure if creating a function for them would save time

notebooks/quick-look/sharing-experiments.ipynb
```diff
@@ -115,26 +115,86 @@
...
-      "\tid: c1b6cb3a-0ad1-4932-914d-ba53a054891b, parameters: [('strategy', 'median'), ('n_neighbors', 10)], metrics: [('accuracy', 0.7403846153846154)]\n",
-      "\tid: c4277327-381a-4885-aba4-a07c050463a5, parameters: [('strategy', 'most_frequent'), ('n_neighbors', 10)], metrics: [('accuracy', 0.75)]\n",
-      "\tid: ce24eeef-4686-4fc7-8c0a-e73d6c9cdb71, parameters: [('strategy', 'mean'), ('n_neighbors', 15)], metrics: [('accuracy', 0.7596153846153846)]\n",
-      "\tid: d4ea2fe7-061e-4f5e-8958-e6ac29025708, parameters: [('strategy', 'most_frequent'), ('n_neighbors', 15)], metrics: [('accuracy', 0.7596153846153846)]\n",
-      "\tid: d9fe2005-824c-4e23-9809-e0459e57d78a, parameters: [('strategy', 'most_frequent'), ('n_neighbors', 20)], metrics: [('accuracy', 0.7211538461538461)]\n",
-      "\tid: f497245a-6149-4604-9ceb-da74ae9855d4, parameters: [('strategy', 'median'), ('n_neighbors', 20)], metrics: [('accuracy', 0.7211538461538461)]\n"
+      "\tid: 0049ae3c-bd1b-4fd5-8b2a-f949fa2da015, parameters: [('strategy', 'mean'), ('n_neighbors', 15)], metrics: [('accuracy', 0.6826923076923077)]\n",
```
ryanSoley, line 126: can you delete all your experiments in the `rubicon_root` directory, then rerun the logging quicklook and this notebook so this has the same number of experiments as it did before you edited it?

you've probably re-ran the logging quicklook multiple times, which is why this has so many duplicates

notebooks/quick-look/sharing-experiments.ipynb
```diff
@@ -156,6 +216,79 @@
...
+      "    args:\n",
+      "      experiment_id: 8604604d-f1f0-4d4c-a0d8-61a71bd9d55e\n",
+      "      project_name: Update Catalog Example\n",
+      "      urlpath: ./rubicon-root\n",
+      "    driver: rubicon_ml_experiment\n"
+     ]
+    }
+   ],
+   "source": [
+    "import intake\n",
```
ryanSoley, line 261: we already imported intake in the cell before - no need to do it again

ryanSoley (commented):
just ran the notebook - few more things, mainly to clear some stuff up in the example that will be rendered in the docs

ryanSoley (approved):
awesome 🎉

Issue #317, opened 2022-12-06
tag display in experiments table


ryanSoley:
**Describe the bug**

when an experiment with multiple tags (say ["tag_a", "tag_b"]) is displayed in the experiments table, they show up as "tag_atag_b"

**Steps/Code to reproduce bug**

run the code in this example, but change `experiment = project.log_experiment()` to `experiment = project.log_experiment(tags=["tag_a", "tag_b"])`, then take a look at the tags column in the experiments table
> https://capitalone.github.io/rubicon-ml/visualizations/experiments-table.html

**Expected behavior**

the experiments table should display "tag_a, tag_b" or something like that

PR #318, opened 2022-12-06
tag bug fix in experiment table

Tejass9922:
closes: #317 

---

## What
  * Bug fix for the way tags were being displayed in the experiment table
  * Tags are now comma separated with a space character in the table

## How to Test
  * "run the code in this example, but change experiment = project.log_experiment() to experiment = 
     project.log_experiment(tags=["tag_a", "tag_b"]), then take a look at the tags column in the experiments table"
     https://capitalone.github.io/rubicon-ml/visualizations/experiments-table.html


ryanSoley (approved):
thanks!

PR #321, opened 2022-12-09
encode/decode `numpy` objects

ryanSoley:
## What
  * adds support for `numpy` generics and arrays to the JSON serializer

## How to Test
  * run the new tests: `python -m pytest tests/unit/repository/utils/test_json.py`
  * try logging some parameters or metrics with `numpy` types:

```python
experiment.log_metric(name="int32", value=np.int32(5))
experiment.log_metric(name="ndarray", value=np.array([np.float32(0.5)]*4).reshape((2, 2)))
```

### DIFF
--- a/rubicon_ml/repository/utils/json.py
+++ b/rubicon_ml/repository/utils/json.py
@@ -1,6 +1,9 @@
 import dataclasses
 import json
+from base64 import b64decode, b64encode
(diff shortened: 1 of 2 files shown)


armaan-dhillon (approved):
Nice!

Issue #313, opened 2022-12-01
`RubiconJSON` CLI integration


ryanSoley:
**Is your feature request related to a problem? Please describe**

we should be able to query rubicon experiments from the command line without using python directly

**Describe the solution you'd like**

a new subcommand should be added to the CLI to enable querying via the `RubiconJSON` class. something like:

`rubicon_ml search --root-dir ./rubicon-root --project-name "my project" "$..experiment[*].metric"`

this should output the same JSON to the command line as it would in the [notebook example](https://github.com/capitalone/rubicon-ml/blob/jsonpath-poc/notebooks/dev/jsonpath.ipynb)

**Additional context**

we use [the `click` library](https://click.palletsprojects.com/en/8.1.x/) for our CLI. don't worry about duplicating code between this new subcommand and the existing ui one - I think we're ultimately gonna deprecate and remove the ui subcommand

this relies on #306 and #312 being completed first

PR #328, opened 2022-12-16
Rubicon json cli integration

Tejass9922:
closes: #313  

---

## What
  *  Adds functionality to retrieve rubicon experiments via the CLI
  * CLI takes in `project-name` `root-dir` and `color` as optional inputs
  * Gives users the ability to set `root-dir` and `project-name` as environment variables for future reference
  * Testing
  

## How to Test
  * run pytest 
  * run  snippet on CLI as sample `rubicon_ml search --root-dir ./rubicon-root --project-name "my project" "$..experiment[*].metric"`

### DIFF
--- a/rubicon_ml/cli.py
+++ b/rubicon_ml/cli.py
@@ -1,3 +1,4 @@
+import os
 import warnings
 
 import click
@@ -87,27 +88,70 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
 @click.option(
     "--root-dir",
     type=click.STRING,
+    default=os.environ.get("RUBICON_ROOT_DIR"),
     help="The absolute path to the top level folder holding the rubicon-ml project.",
-    required=True,
+    required=False,
 )
 @click.option(
-    "--project_name",
+    "--project-name",
+    default=os.environ.get("RUBICON_PROJECT_NAME"),
     type=click.STRING,
     help="Name of rubicon project to retrieve experiments from",
-    required=True,
+    required=False,
+)
+@click.option(
+    "--color",
+    "-c",
+    type=click.STRING,
+    help="Color of query output",
+    required=False,
 )
 @click.argument(
     "query",
     nargs=1,
     required=True,
 )
-def json(project_name, query):
-    rb_json = RubiconJSON(projects=project_name)
-    experiments = rb_json.search(query)
-    return experiments
+def search(root_dir, project_name, color, query):
+    """
+    This function provides capability to query  rubicon experiments from the command line without using python
+    directly and output the JSON in the command line.
+    """
+
+    proj_name_env = "RUBICON_PROJECT_NAME"
+    root_dir_env = "RUBICON_ROOT_DIR"
+    fg_color = "bright_yellow"
+
+    if (proj_name_env not in os.environ and project_name is None) or (
+        root_dir_env not in os.environ and root_dir is None
+    ):
+        warnings.warn(
+            "No previous project name and/or Rubicon root directory saved, enter a new one"
+        )
+        return
+
+    if not project_name:
+        project_name = os.environ["RUBICON_PROJECT_NAME"]
+    if not root_dir:
+        root_dir = os.environ["RUBICON_ROOT_DIR"]
+
+    rubicon = Rubicon(persistence="filesystem", root_dir=root_dir)
+    try:
+        project = rubicon.get_project(name=project_name)
+    except Exception:
+        warnings.warn(
+            "Rubicon project not found with given name, try again with a different project name"
+        )
+        return
+
+    rb_json = RubiconJSON(projects=project)
+    res = rb_json.search(query)
+
+    if "color" in click.get_current_context().params:
+        fg_color = color
+
+    click.secho(res, fg=fg_color)
 
 
 # CLI groups
-
 if __name__ == "__main__":
     cli()
--- a/rubicon_ml/cli.py
+++ b/rubicon_ml/cli.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import warnings
 
 import click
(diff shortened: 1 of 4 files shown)


Tejass9922:
Need to add testing still

rubicon_ml/cli.py
```diff
@@ -78,7 +80,78 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+    This function provides capability to query  rubicon experiments from the command line without using python
+    directly and output the JSON in the command line.
+    """
+
+    proj_name_env = "RUBICON_PROJECT_NAME"
+    root_dir_env = "RUBICON_ROOT_DIR"
+    fg_color = "bright_yellow"
+
+    if (proj_name_env not in os.environ and project_name is None) or (
+        root_dir_env not in os.environ and root_dir is None
```
ryanSoley, line 125: I think the first part of each of the and clauses here are unnecessary, right? like if `proj_name_env` isn't in `os.environ` then `project_name` will be `None` since `os.environ.get("RUBICON_PROJECT_NAME")` returned `None` for the default - so we can just check `if project_name is None` and it'll achieve the same thing

rubicon_ml/cli.py
```diff
@@ -78,7 +80,78 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+    ):
+        warnings.warn(
+            "No previous project name and/or Rubicon root directory saved, enter a new one"
```
ryanSoley, line 128:
```suggestion
            "No --root-dir or --project-name provided. Exiting..."
```

tests/unit/test_cli.py
```diff
@@ -32,3 +32,73 @@ def test_cli(mock_get_project, mock_projects, mock_run_server, project_client):
...
+                project.name,
+                "--color",
+                TEST_COLOR,
+                QUERY,
+            ],
+        )
+
+    assert result_a.exit_code == 0
+    assert len(caught_warnings_a) == 0
+    assert "No previous project name and/or Rubicon root directory saved, enter a new one" in str(
```
ryanSoley, line 99:
```suggestion
    assert "No --root-dir or --project-name provided. Exiting..." in str(
```

rubicon_ml/cli.py
```diff
@@ -78,7 +80,78 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+        )
+        return
 
+    if not project_name:
+        project_name = os.environ["RUBICON_PROJECT_NAME"]
+    if not root_dir:
+        root_dir = os.environ["RUBICON_ROOT_DIR"]
```
ryanSoley, line 135: pretty sure this block is redundant as well. click will set `project_name` equal to `os.environ.get("RUBICON_PROJECT_NAME")` via the default. same with the other

also if either of these are `None` we'd never get to this point since the block above this would raise the warning and return

rubicon_ml/cli.py
```diff
@@ -78,7 +80,78 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+    """
+
+    proj_name_env = "RUBICON_PROJECT_NAME"
+    root_dir_env = "RUBICON_ROOT_DIR"
+    fg_color = "bright_yellow"
+
+    if (proj_name_env not in os.environ and project_name is None) or (
+        root_dir_env not in os.environ and root_dir is None
+    ):
+        warnings.warn(
```
ryanSoley, line 127: since we're returning after raising this, I think I'd argue it's more than a warning. based on a quick search it seems like using `click.secho` for the message like you do later on and then `sys.exit` instead of `return` would be the preferred thing here so we can exit with an error code
> https://github.com/pallets/click/issues/747

rubicon_ml/cli.py
```diff
@@ -78,7 +80,78 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+    if not project_name:
+        project_name = os.environ["RUBICON_PROJECT_NAME"]
+    if not root_dir:
+        root_dir = os.environ["RUBICON_ROOT_DIR"]
+
+    rubicon = Rubicon(persistence="filesystem", root_dir=root_dir)
+    try:
+        project = rubicon.get_project(name=project_name)
+    except Exception:
+        warnings.warn(
```
ryanSoley, line 141: same comment as the last warning - `secho` and `sys.exit`

rubicon_ml/cli.py
```diff
@@ -78,7 +80,78 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+        )
+        return
+
+    rb_json = RubiconJSON(projects=project)
+    res = rb_json.search(query)
+
+    if "color" in click.get_current_context().params:
+        fg_color = color
+
+    click.secho(res, fg=fg_color)
```
ryanSoley, line 152: good find!

rubicon_ml/cli.py
```diff
@@ -78,7 +80,78 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+        project_name = os.environ["RUBICON_PROJECT_NAME"]
+    if not root_dir:
+        root_dir = os.environ["RUBICON_ROOT_DIR"]
+
+    rubicon = Rubicon(persistence="filesystem", root_dir=root_dir)
+    try:
+        project = rubicon.get_project(name=project_name)
+    except Exception:
+        warnings.warn(
+            "Rubicon project not found with given name, try again with a different project name"
```
ryanSoley, line 142: if you `except Exception as e` you can just have this string be the error's original message (its either `str(e)` or `e.msg` or something like that)

ryanSoley (commented):
looking good! I think there's a handful of extra stuff re: default handling in here, but lmk if I'm misunderstanding how the click defaults work

rubicon_ml/cli.py
```diff
@@ -78,7 +81,63 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
 
+    rubicon = Rubicon(persistence="filesystem", root_dir=root_dir)
+    try:
+        project = rubicon.get_project(name=project_name)
+    except Exception as e:
+        click.secho(e, fg="red")
+        sys.exit()
+
+    rb_json = RubiconJSON(projects=project)
+    res = rb_json.search(query)
```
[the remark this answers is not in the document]
ryanSoley: actually `res` can be a list, so it is gonna be more than just `res.match`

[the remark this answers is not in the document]
ryanSoley: and also its `res.value`, not `match` - my bad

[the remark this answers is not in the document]
ryanSoley: double check this, but I think adding this after line 133 should do the trick
```
if not isinstance(res, list):
    res = [res]
matches = [r.value for r in res]
```

ryanSoley (approved):

Issue #330, opened 2022-12-20
return rubicon objects with proper parents


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

the rubicon objects that can optionally be returned by `RubiconJSON` are currently utilizing a `NoOpParent` instead of the proper experiment/project parent
> https://github.com/capitalone/rubicon-ml/blob/jsonpath-poc/rubicon_ml/client/rubicon_json.py#L24

this means that any operations that need to actually reach out to the filesystem will not be possible, essentially making the objects read-only

**Describe the solution you'd like**

we should try to associate the proper parents with each returned object so that they will be fully functional rubicon objects

this will likely require the inspection of the `match.value` objects returned from each query. the JSON in `match.value` should have an `experiment_id` (feature, metric, parameter), or a `parent_id` (experiment, artifact, dataframe)

projects are the exception - they require a config object rather than a parent. ‼️ **this is actually incorrect in the current implementation and I didn't catch it before merging to the integration branch, so we'll have to address it too** ‼️ 

there are a few steps that'll be required in the process:
* for each match in the results object:
  * extract the `experiment_id` or `parent_id` from the result
  * if the queried object is not a project:
    * if `RubiconJSON` is instantiated with `experiments` as an input and the parent of the queried object is an experiment, the parent should be in that list
    * if `RubiconJSON` is instantiated with `projects` as an input and the parent of the queried object is a project, the parent should be in that list
    * if `RubiconJSON` is instantiated with `projects` as an input and the parent of the queried object is an experiment, the parent should retrievable from one of those projects using `project.experiment(id=...`
    * if `RubiconJSON` is instantiated with top level rubicon objects as an input, we'll need to leverage `get_project(id=...` and `experiment(id=...` to retrieve the proper parents whether they be a project or experiment
  * if the queried object is a `project`:
    * we must be retrieving it from a top level rubicon input to `RubiconJSON` - so we can just take the config object off that top level rubicon object and dump it into the new project

**Describe alternatives you've considered**

if this ends up being infeasible, its fine to just leave the `NoOpParents`

**Additional context**
Add any other context, code examples, or references to existing implementations about the feature request here.

ryanSoley:
closing for now - can revisit later if necessary

PR #333, opened 2023-01-11
Parent object search fix

Tejass9922:
closes: #330 
---

## What
  * First draft of returning proper parents of the queried objects inside of the `RubiconJSON.search` function
  * There may potentially be some redundancy that needs to be looked over 

## How to Test
  *


rubicon_ml/client/rubicon_json.py
```diff
@@ -214,6 +243,92 @@ def _rubicon_to_json(self, rubicon_objects):
 
         return json
 
+    def _fetch_parent_object(self, queried_object, id, return_type):
+        """
+        Returns the appropriate parent of the queried object to eventually append
+        to the return list of the query
+        """
+        if isinstance(queried_object, Project):
```
ryanSoley, line 251: this will never be true, `match.value` won't ever be a rubicon object

ryanSoley:
looks like the error in the tests is happening when there is a single project or single rubicon object provided to the class's init. the init should probably sanitize the input and make any single objects into lists

rubicon_ml/client/rubicon_json.py
Tejass9922: Good catch, thank you, I'll just use the return_type as an indicator

ryanSoley (commented):
I'm trying to run this notebook from the jsonpath-poc integration branch (I think you need a rebases - doesn't look like the notebook is on your branch) and running into errors. looks like you're passing `return_type` to your `_fetch_query_string_type` and its only expecting `query_string`

rubicon_ml/client/rubicon_json.py
```diff
@@ -76,79 +97,157 @@ def search(self, query, return_type=None):
...
+                            if key in match.value:
+                                del match.value[key]
+                    id = match.value["id"]
+                    parent = self._fetch_parent_object(
+                        queried_object=query_string_type, id=id, return_type="experiment"
+                    )
+                    return_objects.append(Experiment(DomainExperiment(**match.value), parent))
+
                             if key in match.value[i]:
                                 del match.value[i][key]
```
ryanSoley, line 157: I think something got messed up in the merge here. I'm getting a syntax error

ryanSoley:
closing for now in favor of #386 - we can revisit and complete this if the users ask for this additional functionality

Issue #347, opened 2023-03-23
read zipped archives of experiments


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

rubicon writes many files while logging. retrieval of these files from network based filesystems can be slow due to the number of individual files.

**Describe the solution you'd like**

a function `project.experiments_from_archive` that reads and unzips `project`'s archived experiments into the local filesystem

**Additional context**

`project.experiments_from_archive(local_root="/local-rubicon-root")` should read the experiments from all of `project`'s archives in "rubicon-root/project-name/archives/" and write them to "local-rubicon-root/project-name/experiments"  (we'll likely need to copy the project's metadata to the `local_root` too)

`project.experiments_from_archive(local_root="/local-rubicon-root", latest_only=True)` should read the experiments from the `project`'s most recently logged archive in "rubicon-root/project-name/archives/" and write them to "local-rubicon-root/project-name/experiments"

Issue #346, opened 2023-03-23
write zipped archives of experiments


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

rubicon writes many files while logging. retrieval of these files from network based filesystems can be slow due to the number of individual files. 

**Describe the solution you'd like**

a function `project.archive` that zips the given experiments into a single file and writes it to a different filesystem at "rubicon-root/project-name/archives/name.zip"

**Additional context**

`project.archive()` should zip all experiments in the project into a file "rubicon-root/project-name/archives/archive-{timestamp}.zip" where "rubicon-root/" is the current project's root

`project.archive(experiments=my_experiments)` should zip only the experiments (belonging to `project`) in the list `my_experiments` into a file "rubicon-root/project-name/archives/archive-{timestamp}.zip"

eventually, we'll incorporate a `remote_root="s3://my-bucket/my-key"` argument to send the archive to a different filesystem

ryanSoley:
some docs that may be useful:

it may be possible to use `fsspec` directly for zipping the files
> https://filesystem-spec.readthedocs.io/en/latest/index.html

however, it may just be easier to work with the builtin python modules `shutil` and `zipfile`
> https://docs.python.org/3/library/shutil.html#archiving-operations
> https://docs.python.org/3/library/zipfile.html#module-zipfile

ryanSoley:
closed w/ #349

PR #349, opened 2023-04-05
added read/write for experiment archives

armaan-dhillon:
Adresses: #346 and #347 

---

## What
  * Implements archival of logged experiments in a `project` (write)
  * Implements retrieval of archived experiments from a "remote" archive (read)
  * Adds some helpful methods to the `BaseRepository`
  * Adds tests

## How to Test
  * python -m pytest tests/unit/client/test_project_client.py
  * Or run the following cells in a Jupyter notebook (within rubicon-ml-dev environment).
```
```
```
rubiconA = Rubicon(persistence="filesystem", root_dir='rubicon-rootA')
projectA = rubiconA.get_or_create_project("ArchiveTesting")
experiment1 = projectA.log_experiment(name="experiment1")
experiment2 = projectA.log_experiment(name="experiment2")
projectA.archive()
```
```
rubiconB = Rubicon(persistence="filesystem", root_dir='rubicon-rootB')
projectB = rubiconB.get_or_create_project("ArchiveTesting")
```
```
projectB.experiments_from_archive(remote_root="rubicon-rootA")
```
```
! ls './rubicon-rootB/ArchiveTesting'
```
```
! ls './rubicon-rootB/ArchiveTesting/experiments'
```

### DIFF
--- a/rubicon_ml/client/project.py
+++ b/rubicon_ml/client/project.py
@@ -395,47 +395,6 @@ class Project(Base, ArtifactMixin, DataframeMixin):
         else:
             print("zip archive not created")
 
-    # @failsafe
-    # def experiments_from_archive(self, local_root:str, latest_only: Optional[bool]=False):
-    #     if os.path.exists(os.path.join(local_root, self.name)) == False:
-    #         self.repository._mkdir(os.path.join(local_root, self.name))
-
-    #     root_dir = self.repository.root_dir
-    #     shutil.copy(os.path.join(root_dir, self.name,"metadata.json"), os.path.join(local_root, self.name))
-
-    #     archive_dir = os.path.join(root_dir, self.name, "archives")
-    #     dest_experiments_dir = os.path.join(local_root, self.name, "experiments")
-
-    #     if os.path.exists(dest_experiments_dir) == False:
-    #         self.repository._mkdir(dest_experiments_dir)
-
-    #     dest_experiments_dir_mod_time = os.path.getmtime(dest_experiments_dir)
-    #     if os.path.exists(os.path.join(local_root, self.name)) == False:
-    #         self.repository._mkdir(os.path.join(local_root, self.name))
-
-    #     shutil.copy(os.path.join(root_dir, self.name,"metadata.json"), os.path.join(local_root, self.name))
-
-    #     if not latest_only:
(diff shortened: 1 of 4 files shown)


rubicon_ml/client/project.py
```diff
@@ -347,6 +352,84 @@ def dataframes(self, tags=[], qtype="or", recursive=False, name=None):
 
         return self._dataframes
 
+    @failsafe
+    def archive(self, experiment_names: Optional[list[str]] = None):
```
ryanSoley, line 356: I think we should ultimately just take in experiment objects here instead of names or IDs. then users can just get them by name or ID or tags or whatever they want before the send them in

rubicon_ml/client/project.py
```diff
@@ -347,6 +352,84 @@ def dataframes(self, tags=[], qtype="or", recursive=False, name=None):
...
+                    archive.write(file_path, os.path.basename(file_path))
+                archive.close()
+
+        if os.path.exists(zip_archive_filename):
+            print("zip archive created")
+        else:
+            print("zip archive not created")
+
+    @failsafe
+    def experiments_from_archive(self, remote_root: str, latest_only: Optional[bool] = False):
```
ryanSoley, line 399: we'll probably need to replace these `os` calls that interact with the filesystem with `fsspec` calls that do the same thing in order to get it working with S3

rubicon_ml/client/project.py
ryanSoley: we may also want to consider a `remote_root` option similar to `experiments_from_archive`'s for this function as well. that way we could facilitate a workflow thats something more like 

```
write individual experiments locally -> 
archive to S3 -> 
later read archives from S3 back to local and unzip
``` 

instead of 

```
write individual experiments locally -> 
archive locally -> 
sync everything to S3 -> 
later read archives from S3 back to local and unzip
```

ryanSoley (commented):
looking great! we'll make some followup issues to take care of the stuff I mentioned in the comments and use this as a feature branch

the error in the tests just needs you to use `List` from `typing` instead of the builtin `list`

```
def archive(self, experiment_names: Optional[list[str]] = None):
TypeError: 'type' object is not subscriptable
Error: Process completed with exit code 1.
```

Issue #357, opened 2023-05-04
read/write with multiple backends


ryanSoley:
**Is your feature request related to a problem? Please describe**

sometimes it is desirable to write rubicon logs to multiple backends. currently, that'd be accomplished by instantiating a second `Rubicon` object and duplicating any logging code:

```python
rb_a = Rubicon(persistence="filesystem", root_dir="./local/root")
log_stuff(rb_a)  # calls to `log_` functions log to local filesystem

rb_b = Rubicon(persistence="filesystem", root_dir="s3://remote/bucket/root")
log_stuff(rb_b)  # calls to `log_` functions log to remote S3 filesystem
```

**Describe the solution you'd like**

we should create a new repository layer that can configure multiple backends at once for reading and writing. the client level entry point should look something like this:

```python
rb = Rubicon(composite_config=[
    {"persistence": "filesystem", "root_dir": "./local/root"},
    {"persistence": "filesystem", "root_dir": "s3://remote/bucket/root"},
])
log_stuff(rb)  # calls to `log_` functions log to both local filesystem and remote S3 filesystem
```

since the `Rubicon` object exposes arbitrary kwargs as `**storage_options` , we don't need to update the `Rubicon` object
to allow `composite_config` to be passed in. we just need to be aware that we need to retrieve it like this in the config: `self.storage_options.get("composite_config")`

at the repository layer, we'll need to create a new backend repo class (something like `CompositeRepository` in **rubicon_ml/repository/composite.py**). the current repo extensions are generally instantiated with a `root_dir` and `storage_options`, but this new one will be instantiated with other instantiated repos - perhaps something like this

```python
class CompositeRepository(BaseRepository):
    def __init__(self, repositories):
        self.repositories = repositories
        # other init ...
```

then, we'll need to update the [client **config.py**](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/config.py) to handle the `composite_config` argument properly. currently, for a single backend, we set `self.repository` to the single backend in the config's init. in the case of a composite backend, we'll need to create a repository for each of the entries in `composite_config` and append them to a list which can then be passed to `CompositeRepository`, which will ultimately be `self.repository` on the config

[...]

PR #359, opened 2023-05-12
added multiple backends initial code

armaan-dhillon:
closes: #357 

---

## What
  * Adds new repository layer `CompositeRepository` in `rubicon_ml/repository/composite.py` to handle multiple backends
  * Updates `rubicon_ml/client/config.py` to be able to handle `Rubicon` instantiation with multiple/compisite backends
  * Overrides "Filesystem helpers" methods in `CompositeRepository` for reading/writing from/to multiple backends

## How to Test
  *

### DIFF
--- a/rubicon_ml/client/config.py
+++ b/rubicon_ml/client/config.py
@@ -2,7 +2,12 @@ import os
 import subprocess
 
 from rubicon_ml.exceptions import RubiconException
-from rubicon_ml.repository import LocalRepository, MemoryRepository, S3Repository
+from rubicon_ml.repository import (
+    CompositeRepository,
+    LocalRepository,
+    MemoryRepository,
+    S3Repository,
+)
 
 
 class Config:
@@ -38,12 +43,22 @@ class Config:
     def __init__(
         self, persistence=None, root_dir=None, is_auto_git_enabled=False, **storage_options
     ):
-        self.persistence, self.root_dir, self.is_auto_git_enabled = self._load_config(
-            persistence, root_dir, is_auto_git_enabled
-        )
-        self.storage_options = storage_options
+        composite_config = self.storage_options.get("composite_config", None)
+        if composite_config is None:
+            self.persistence, self.root_dir, self.is_auto_git_enabled = self._load_config(
+                persistence, root_dir, is_auto_git_enabled
+            )
+            self.storage_options = storage_options
 
-        self.repository = self._get_repository()
+            self.repository = self._get_repository()
+        else:
+            repositories = []
+            for config in composite_config:
+                self.persistence, self.root_dir, self.is_auto_git_enabled = self._load_config(
+                    config["persistence"], config["root_dir"], config["is_auto_git_enabled"]
+                )
+            repositories.append(self._get_repository())
+            self.repository = CompositeRepository(repositories)
 
     def _check_is_in_git_repo(self):
         """Raise a `RubiconException` if not called from within a `git` repository."""
--- a/rubicon_ml/repository/__init__.py
+++ b/rubicon_ml/repository/__init__.py
@@ -1,10 +1,12 @@
 from rubicon_ml.repository.base import BaseRepository
+from rubicon_ml.repository.composite import CompositeRepository
 from rubicon_ml.repository.local import LocalRepository
 from rubicon_ml.repository.memory import MemoryRepository
 from rubicon_ml.repository.s3 import S3Repository
 
 __all__ = [
     "BaseRepository",
+    "CompositeRepository",
     "LocalRepository",
     "MemoryRepository",
     "S3Repository",
--- a/rubicon_ml/repository/composite.py
+++ b/rubicon_ml/repository/composite.py
@@ -1,0 +1,159 @@
+import os
+import warnings
+
+from rubicon_ml.exceptions import RubiconException
+from rubicon_ml.repository import BaseRepository
+from rubicon_ml.repository.utils import json
+
+
+class CompositeRepository(BaseRepository):
+    def __init__(self, repositories):
+        self.repositories = repositories
+
+    # --- Filesystem Helpers ---
+
+    def _cat(self, path):
+        """Returns the contents of the file at `path`."""
+        for repo in self.repositories:
+            try:
+                result = repo._cat(path)
+            except Exception as err:
+                last_err = err
+                pass
+            else:
+                return result
+        raise RubiconException(last_err)
+
+    def _cat_paths(self, metadata_paths):
+        """Cat `metadata_paths` to get the list of files to include.
+        Ignore FileNotFoundErrors to avoid misc file errors, like hidden
+        dotfiles.
+        """
+        for repo in self.repositories:
+            try:
+                files = []
+                for path, metadata in repo.filesystem.cat(
+                    metadata_paths, on_error="return"
+                ).items():
+                    if isinstance(metadata, FileNotFoundError):
+                        warning = f"{path} not found. Was this file unintentionally created?"
+                        warnings.warn(warning)
+                    else:
+                        files.append(metadata)
+            except Exception as err:
+                last_err = err
+                pass
+            else:
+                return files
+        raise RubiconException(last_err)
+
+    def _exists(self, path):
+        """Returns True if a file exists at `path`, False otherwise."""
+        for repo in self.repositories:
+            try:
+                exists = repo.filesystem.exists(path)
+            except Exception as err:
+                last_err = err
+                pass
+            else:
+                return exists
+        raise RubiconException(last_err)
+
+    def _glob(self, globstring):
+        """Returns the names of the files matching `globstring`."""
+        for repo in self.repositories:
+            try:
+                glob = repo.filesystem.glob(globstring, detail=True)
+            except Exception as err:
+                last_err = err
+                pass
+            else:
+                return glob
+        raise RubiconException(last_err)
+
+    def _ls_directories_only(self, path):
+        """Returns the names of all the directories at path `path`."""
+        for repo in self.repositories:
+            try:
+                directories = [
+                    os.path.join(p.get("name"), "metadata.json")
+                    for p in repo.filesystem.ls(path, detail=True)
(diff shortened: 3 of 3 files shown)


rubicon_ml/client/config.py
```diff
@@ -38,12 +38,24 @@ class Config:
     def __init__(
         self, persistence=None, root_dir=None, is_auto_git_enabled=False, **storage_options
     ):
-        self.persistence, self.root_dir, self.is_auto_git_enabled = self._load_config(
-            persistence, root_dir, is_auto_git_enabled
-        )
-        self.storage_options = storage_options
+        if storage_options is not None and storage_options.get("composite_config") is not None:
```
ryanSoley, line 41:
```suggestion
        if storage_options is not None and "composite_config" in storage_options:
```

rubicon_ml/client/config.py
```diff
@@ -38,12 +38,24 @@ class Config:
...
+            composite_config = storage_options.get("composite_config")
+            self.storage_options = storage_options
```
ryanSoley, line 43: this is duplicated on line 51 - we can move it outside the if/else

rubicon_ml/client/config.py
```diff
@@ -38,12 +38,24 @@ class Config:
...
+            repositories = []
+            for config in composite_config:
+                self.persistence, self.root_dir, self.is_auto_git_enabled = self._load_config(
```
ryanSoley, line 46: are `self.persistence` and `self.root_dir` accessed at all outside of `self._get_repository`? it seems like `self.persistence` is used to choose the repo then `self.root_dir` gets set on the repo then is always accessed as `repo.root_dir` afterwards. if so we should refactor later

rubicon_ml/client/dataframe.py
```diff
@@ -43,14 +43,22 @@ def get_data(self, df_type="pandas"):
...
+            try:
+                self._data = repo.get_dataframe_data(
+                    project_name,
+                    self.id,
+                    experiment_id=experiment_id,
+                    df_type=df_type,
+                )
+            except Exception as err:
+                return_err = err
+                pass
```
ryanSoley, line 57: we don't need the `pass`es in all of these - I just had it in the example in the issue cause there wasn't any other code in the `except` block

tests/integration/test_dataframe_logging.py
```diff
@@ -59,8 +59,9 @@ def test_df_read_error(rubicon_local_filesystem_client):
 
     # simulate user forgetting to set `df_type` to `dask` when reading a logged dask df
     with pytest.raises(RubiconException) as e:
-        read_dataframe.get_data()
-
+        res = read_dataframe.get_data()
+        print(res)
+    # pdb.set_trace()
```
ryanSoley, line 64: looks like these are left over from debugging

rubicon_ml/repository/base.py
```diff
@@ -83,7 +83,7 @@ def _ls(self, path):
 
     def _mkdir(self, dirpath):
         """Creates a directory `dirpath` with parents."""
-        return self.filesystem.mkdir(dirpath, parents=True, exist_ok=True)
+        return self.filesystem.mkdirs(dirpath, exist_ok=True)
```
ryanSoley, line 86: can you explain why we're changing this here just in case we need to come back and check why

rubicon_ml/client/config.py
armaan-dhillon: `self._get_repository()` calls `self._get_protocol` so they're getting used there again. Seems like the factoring is okay

rubicon_ml/repository/base.py
armaan-dhillon: Sure! So `fsspec`'s `mkdir` command takes in a `create_parents` kwarg and not a `parents` kwarg, and at the same time it does not take in an `exists_ok` kwarg which we need it too. `fsspec`'s `mkdirs` command, however, does take in an `exist_ok` kwarg and it is noted in the `mkdir` method docs that if `create_parents = True` then 'mkdir' is equivalent to 'mkdirs' which will "create parents" like we want without needing to pass in a `create_parents` kwarg. So we make the change to `mkdirs` instead of `mkdir`.

rubicon_ml/client/project.py
```diff
@@ -289,10 +290,17 @@ def experiment(self, id=None, name=None):
                 )
 
             experiment = experiments[-1]
```
ryanSoley, line 292: lines 281-293 need to be inside your try/except as well. this first block is for getting an experiment by name - right now, only retrieval by ID will work with the mult-backend. this whole `if` should be inside the `try`. then there should also be an `else` in the `try` that has the current `try` code

rubicon_ml/client/artifact.py
```diff
@@ -40,10 +41,18 @@ def __init__(self, domain, parent):
...
+            try:
+                self._data = repo.get_artifact_data(
+                    project_name, self.id, experiment_id=experiment_id
+                )
+            except Exception as err:
+                return_err = err
+            else:
+                return
+        if self._data is None:
+            return RubiconException(return_err)
```
ryanSoley, line 55: this needs to be a `raise`

rubicon_ml/client/artifact.py
```diff
@@ -58,11 +67,16 @@ def get_data(self, unpickle=False):
...
+        for repo in self.repositories:
+            try:
+                data = repo.get_artifact_data(project_name, self.id, experiment_id=experiment_id)
+            except Exception as err:
+                return_err = err
+            else:
+                if unpickle:
+                    data = pickle.loads(data)
+                return data
+        return RubiconException(return_err)
```
ryanSoley, line 79: this needs to be a `raise`

rubicon_ml/client/artifact.py
ryanSoley: can we also change all of the `raise RubiconException(return_err)` to `raise RubiconException("all configured storage backends failed") from return_err`?

using the `from` syntax will preserve the original error's callback when its printed, which should be easier for users to debug

rubicon_ml/client/mixin.py
```diff
@@ -259,13 +261,20 @@ def artifact(self, name=None, id=None):
                 )
 
             artifact = artifacts[-1]
+            return artifact
```
ryanSoley, line 264: same comment as with the experiment getting function - this first part needs to be inside the try too to handle getting by name and ID

rubicon_ml/client/mixin.py
```diff
@@ -382,16 +396,23 @@ def dataframe(self, name=None, id=None):
                 )
 
             dataframe = dataframes[-1]
+            return dataframe
```
ryanSoley, line 399: same comment as with the experiment and artifact getting function - this first part needs to be inside the try too to handle getting by name and ID

rubicon_ml/client/project.py
```diff
@@ -379,11 +393,23 @@ def archive(self, experiments: Optional[List[Experiment]] = None, remote_rubicon
             if not isinstance(remote_rubicon, Rubicon):
                 raise ValueError("`remote_rubicon` must be of type `rubicon_ml.client.Rubicon`")
             else:
-                return self.repository._archive(
-                    self.name, experiments, remote_rubicon.repository.root_dir
-                )
+                res = []
```
ryanSoley, line 396: lets remove these changes to the archiving and leave it as-is for now. if a user archives a multi-repo to one s3 bucket, we don't need to archive it more than once - it'd just keep overwriting itself with the same data

rubicon_ml/client/rubicon.py
```diff
@@ -120,12 +128,18 @@ def get_project(self, name=None, id=None):
...
+                    return_err = err
+                else:
+                    project = Project(project, self.config)
+                    return project
         else:
             project = [p for p in self.projects() if p.id == id][0]
-
-        return project
+            return project
+        raise RubiconException(return_err)
```
ryanSoley, line 142: this is similar to the experiment/dataframe/artifact getter issue except this time you only put the try/except in the first block instead of only in the second one. it should have a similar fix

ryanSoley (commented):
I took a more thorough look - there's a few things, but two main ones
* let's `raise RubiconException() from return_err` instead of `raise RubiconException(return_err)` (further explained below)
  * there are also a couple of errors still being returned, not raised
* there's a problem with the try/except and if/else structure in the single item getters
  * each of them have an if/else that either gets the item by ID or name. in each of your functions, you only put one part of that if/else inside the try/excepts. both parts of those if/else statements need to be inside the try blocks in those functions

armaan-dhillon:
> I took a more thorough look - there's a few things, but two main ones
> 
> * let's `raise RubiconException() from return_err` instead of `raise RubiconException(return_err)` ([further explained below](https://github.com/capitalone/rubicon-ml/pull/359#discussion_r1228556975))
>   
>   * there are also a couple of errors still being returned, not raised
> * there's a problem with the try/except and if/else structure in the single item getters
>   
>   * each of them have an if/else that either gets the item by ID or name. in each of your functions, you only put one part of that if/else inside the try/excepts. both parts of those if/else statements need to be inside the try blocks in those functions

Connected over Zoom, and concluded that the second bullet is not needed since the plural methods (`metrics`, `parameters`, `features`, etc) all have their own `try/except` structures being utilized and the singular methods (`metric`, `parameter`, `feature`, etc) that call those plural methods outside their `try/except` structures don't need to shift the `else`  of when `name is None` case within their own `try` statements.

ryanSoley (commented):
one last thing - looks like some change made the integration test start failing again. let's get that passing. you can run it locally with `python -m pytest -m write_files` (the second one that writes to s3 is still expected to fail, but the other two shouldn't)

you'll just need to update the part at the very beginning and very end that call `rubicon.repository` to use a `for repository in rubicon.repositories` loop, then just `repository` instead of `rubicon.repository` like a lot of the other functions you updated in the client

ryanSoley (approved):
awesome job on this @armaan-dhillon! I'm super excited to get this in

we'll need to make a few followup stories, mostly around more robust testing - especially around the failure cases

Issue #360, opened 2023-05-30
archive to s3


ryanSoley:
**Describe the bug**

archiving to s3 does not currently work. files are only being archived locally, even in the s3 case

**Steps/Code to reproduce bug**

```python
rubicon_a = Rubicon(persistence="filesystem", root_dir="./local/root/")
rubicon_b = Rubicon(persistence="filesystem", root_dir="s3://remote/root/")

project_a = rubicon_a.create_project(name="test")
experiment_a = project_a.log_experiment()

project_a.archive(remote_rubicon=rubicon_b)
```

this results in a new local directory `s3/remote/root` rather than actually archiving to s3

**Expected behavior**

the project and experiment logged by `rubicon_a` should be zipped and written to the s3 bucket, not a different local filesystem

**Additional context**

the solution will likely require two steps:
1. if `remote_rubicon` is provided and points to a local filesystem, write the zipfile locally to the given remote as it currently is done - **or** - if `remote_rubicon` is provided and points to an s3 filesystem, write the zipfile locally to a temporary file
2. if `remote_rubicon` is provided and points to an s3 filesystem, use the `remote_rubicon`'s repository to write the local, temporary, zipped archive to s3 before the temp file or directory is cleaned up (no step 2 required for local - the original was written to the right spot)

you can use pythons built-ins for dealing with temporary files and directories
> https://docs.python.org/3/library/tempfile.html#tempfile-examples

PR #365, opened 2023-06-12
S3 archive bug fix

Tejass9922:
closes: #360

---

## What
  * Fixes bug in archiving to remote rubicon S3. 
  * Added unit test 

## How to Test
  * run unit tests
  * For a more thorough test, utilize a Kubeflow notebook that has an IAM role attached and run archiving functionality against any given S3 bucket attached to the role


rubicon_ml/repository/base.py
```diff
@@ -296,8 +298,10 @@ def _archive(
         -------
         filepath of newly created archive
         """
+        remote_s3 = True if remote_rubicon_root and remote_rubicon_root.startswith("s3") else False
         if remote_rubicon_root is not None:
             archive_dir = os.path.join(remote_rubicon_root, slugify(project_name), "archives")
+
```
ryanSoley, line 304: can we similarly do `root_dir = remote_rubicon_root if remote_rubicon_root is not None else self.root_dir` then just have the `archive_dir = ...` line once?

Issue #142, opened 2021-11-02
add test for the value error in `project`, `metric`, `experiment` and the other getters 


andreafehrman:
**Describe the solution you'd like**
Create a test function so that the value error is correctly thrown in each getter for making sure `name` or `id` is sent as a parameter. 

**Describe alternatives you've considered**
Alternatively, different functions to test this value error in each getter function could be created but it would most likely be more efficient to write one to be used for all.

ryanSoley:
> it would most likely be more efficient to write one to be used for all

I definitely agree - if you wanna take it a step further, you could add something like a `_validate_only_one` helper to the [base client class](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/base.py) that contains an abstracted version of this check
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/client/experiment.py#L97

then you could swap all the duplicates of the lines I just linked with `self._validate_only_one(name=name, id=id)` and write one test for `_validate_one_of`

PR #371, opened 2023-06-21
Add tests for ValueError

stephenpardy:
closes: #142 

---

## What
  *  Adds unit tests to verify that passing either both `name` and `id` as `None` or both as not `None` results in a ValueError.

## How to Test
  * This PR adds tests only which will be triggered during the CI/CD process.


ryanSoley (approved):
looks great! thanks!

ryanSoley (approved):

Issue #368, opened 2023-06-20
Add mypy type hints


stephenpardy:
**Is your enhancement request related to a problem? Please describe**
Currently, Rubicon does not have added mypy type hints which makes type-checking modules that _use_ Rubicon to be less robust.

**Describe the solution you'd like**
Add consistent type hints throughout the Rubicon package, starting with any user-facing functions and classes.

**Describe alternatives you've considered**
Business as usual.

**Additional context**
Type hints will also allow us to add type checks during the CI/CD process that can catch errors.

PR #379, opened 2023-09-10
Add typehints to much of the core code.

stephenpardy:
Partially addresses #368 

Adds partial type hints throughout the code. 
Leaves almost all logic untouched. One exception are areas where the default argument was a list. This has been changed to `None` with the list added in the function body. This is to avoid the mutable default argument issue.

---

## What
  * Adds type hints throughout the code

## How to Test
  * Relying on existing unit tests.
  * Can run `mypy` - will still show around 100 issues.

### DIFF
--- a/rubicon_ml/client/base.py
+++ b/rubicon_ml/client/base.py
@@ -4,6 +4,7 @@ from typing import List, Optional, TYPE_CHECKING
 if TYPE_CHECKING:
     from rubicon_ml.client import Config
     from rubicon_ml.domain import DOMAIN_TYPES
+    from rubicon_ml.repository import BaseRepository
 
 
 class Base:
@@ -25,11 +26,11 @@ class Base:
         return self._domain.__str__()
 
     @property
-    def repository(self) -> Optional[str]:
+    def repository(self) -> Optional[BaseRepository]:
         return self._config.repository if self._config is not None else None
 
     @property
-    def repositories(self) -> Optional[List[str]]:
+    def repositories(self) -> Optional[List[BaseRepository]]:
         if self._config is None:
             return None
 
--- a/rubicon_ml/client/base.py
+++ b/rubicon_ml/client/base.py
@@ -1,3 +1,11 @@
+from typing import List, Optional, TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from rubicon_ml.client import Config
+    from rubicon_ml.domain import DOMAIN_TYPES
+
+
 class Base:
     """The base object for all top-level client objects.
 
@@ -9,19 +17,22 @@ class Base:
         The config, which injects the repository to use.
     """
 
-    def __init__(self, domain, config=None):
+    def __init__(self, domain: DOMAIN_TYPES, config: Optional[Config] = None):
         self._config = config
         self._domain = domain
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self._domain.__str__()
 
     @property
-    def repository(self):
-        return self._config.repository
+    def repository(self) -> Optional[str]:
+        return self._config.repository if self._config is not None else None
 
     @property
-    def repositories(self):
+    def repositories(self) -> Optional[List[str]]:
+        if self._config is None:
+            return None
+
         if hasattr(self._config, "repositories"):
             return self._config.repositories
         else:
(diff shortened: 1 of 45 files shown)


rubicon_ml/client/base.py
```diff
@@ -9,19 +19,22 @@ class Base:
         The config, which injects the repository to use.
     """
 
-    def __init__(self, domain, config=None):
+    def __init__(self, domain: DOMAIN_TYPES, config: Optional[Config] = None):
```
ryanSoley, line 22: I have to look into why the `config` is optional - that doesn't seem right

rubicon_ml/client/config.py
```diff
@@ -29,18 +35,22 @@ class Config:
     """
 
     PERSISTENCE_TYPES = ["filesystem", "memory"]
-    REPOSITORIES = {
+    REPOSITORIES: Dict[str, Type[BaseRepository]] = {
```
ryanSoley, line 38: why do we need the `Type` here?

rubicon_ml/client/mixin.py
```diff
@@ -47,14 +60,14 @@ def _validate_data(self, data_bytes, data_file, data_object, data_path, name):
     @failsafe
     def log_artifact(
         self,
-        data_bytes=None,
+        data_bytes: Optional[bytes] = None,
         data_file=None,
```
ryanSoley, line 64: missed one?

rubicon_ml/client/mixin.py
```diff
@@ -298,8 +316,12 @@ def delete_artifacts(self, ids):
 class DataframeMixin:
     """Adds dataframe support to a client object."""
 
+    _domain: DOMAIN_TYPES
+
     @failsafe
-    def log_dataframe(self, df, description=None, name=None, tags=[]):
+    def log_dataframe(
+        self, df: Union[pd.DataFrame, dd.DataFrame], description=None, name=None, tags=[]
```
ryanSoley, line 323: missed the last couple

rubicon_ml/client/parameter.py
```diff
@@ -22,36 +31,36 @@ class Parameter(Base, TagMixin):
...
 
     @property
-    def name(self):
+    def name(self) -> Optional[str]:
         """Get the parameter's name."""
         return self._domain.name
 
     @property
-    def value(self):
+    def value(self) -> Optional[Union[object, float]]:
```
ryanSoley, line 49: isn't `Union[object, float]` == `object`?

rubicon_ml/client/parameter.py
```diff
@@ -22,36 +31,36 @@ class Parameter(Base, TagMixin):
...
         """Get the parameter's value."""
-        return self._domain.value
+        return getattr(self._domain, "value", None)
```
ryanSoley, line 51: why is this necessary? a parameter domain should never not have a `value` key

rubicon_ml/client/rubicon.py
```diff
@@ -32,7 +33,11 @@ class Rubicon:
     """
 
     def __init__(
-        self, persistence="filesystem", root_dir=None, auto_git_enabled=False, **storage_options
+        self,
+        persistence: Optional[str] = "filesystem",
+        root_dir=None,
+        auto_git_enabled=False,
```
ryanSoley, line 39: missed a couple

ryanSoley (commented):
this is awesome! thanks for getting us started down this path! 

I have a few general questions and also noted a few functions were you missed a variable or two - lmk if any of those omissions were intentional

rubicon_ml/client/base.py
stephenpardy: Yeah, the default value here is `None`, do we need to make it required?

rubicon_ml/client/parameter.py
stephenpardy: Good catch. This was because I didn't hint the `_domain` value properly.

rubicon_ml/client/base.py
ryanSoley: not in this PR, maybe in the future. I think its a remnant of the fact that `Rubicon` used to extend the `Base` and didn't take in a config since it generated it. I'll verify that's the case and if so make an issue to fix it

rubicon_ml/client/config.py
ryanSoley: ahh gotcha, makes sense

rubicon_ml/client/mixin.py
stephenpardy: Thanks.

ryanSoley (approved):
looks great!

PR #384, opened 2023-09-22
multi-backend errors

ryanSoley:
## What
  * only raise the "all configured backends failed" error when there is more than one backend
  * move error handling logic to `Base`

## How to Test
  * `python -m pytest`

### DIFF
--- a/rubicon_ml/client/rubicon.py
+++ b/rubicon_ml/client/rubicon.py
@@ -143,18 +143,18 @@ class Rubicon:
 
         if name is not None:
             return_err = None
+
             for repo in self.repositories:
                 try:
                     project = repo.get_project(name)
                 except Exception as err:
                     return_err = err
                 else:
-                    project = Project(project, self.config)
-                    return project
-            raise RubiconException("all configured storage backends failed") from return_err
+                    return Project(project, self.config)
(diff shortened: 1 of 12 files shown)


rubicon_ml/client/rubicon.py
```diff
@@ -143,18 +143,21 @@ def get_project(self, name=None, id=None):
...
                 else:
-                    project = Project(project, self.config)
-                    return project
-            raise RubiconException("all configured storage backends failed") from return_err
+                    return Project(project, self.config)
+
+            if len(self.repositories) > 1:
+                raise RubiconException("all configured storage backends failed") from return_err
+            else:
+                raise return_err
```
stephenpardy, line 158: Should this also use the new method?

ryanSoley: `Rubicon` doesn't extend from `Base` (even though it reimplements nearly everything on it) - I'm gonna open an issue to look into that

stephenpardy (approved):

rubicon_ml/client/rubicon.py
ryanSoley: #387

Issue #315, opened 2022-12-01
add example & docs for `RubiconJSON`


ryanSoley:
**Is your documentation request related to a problem? Please describe**

since this is a completely new feature, we'll need a new section in the docs for it

**Describe the solution you'd like**

design a notebook showing how the new `RubiconJSON` class works (maybe just adapt [the poc one](https://github.com/capitalone/rubicon-ml/blob/jsonpath-poc/notebooks/dev/jsonpath.ipynb) we're basing this all off) for the documentation. add said notebook to the docs. add any new, public python methods to the [API reference](https://capitalone.github.io/rubicon-ml/api_reference.html)

PR #386, opened 2023-09-22
support JSONpath queries

ryanSoley:
closes: #315 
---

## What
  * finally getting around to merging this integration branch to support JSONpath queries of logged rubicon data
    * take a look at the closed stores in [this milestone](https://github.com/capitalone/rubicon-ml/milestone/8?closed=1) for more context

## To Do
- [x] merge #384 for tests here to pass
- [x] update notebook example
- [x] update PR description and remove draft status

## How to Test
  * `python -m pytest`


rubicon_ml/cli.py
```diff
@@ -78,7 +82,75 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+)
+@click.argument(
+    "query",
+    nargs=1,
+    required=True,
+)
+def search(root_dir, project_name, color, bf, query):
+    """
+    This function provides capability to query  rubicon experiments from the command line without using python
+    directly and output the JSON in the command line.
```
shania-m, line 127: can we get doc strings for the parameters? Or `bf` specifically

rubicon_ml/cli.py
```diff
@@ -78,7 +82,75 @@ def ui(root_dir, page_size, project_name, host, port, debug, storage_options):
...
+
+    rb_json = RubiconJSON(projects=project)
+    res = rb_json.search(query)
+    if not isinstance(res, list):
+        res = [res]
+    matches = [r.value for r in res]
+    if "color" in click.get_current_context().params:
+        fg_color = color
+
+    result = matches if bf else pprint.pformat(matches, indent=1)
```
shania-m, line 149: what is `bf`?

stephenpardy (commented):
We don't currently enforce type hints, but can we please add them here wherever possible to facilitate adding mypy in the future? Thanks!

rubicon_ml/cli.py
ryanSoley: that and the color option were for formatting the output. I replaced `bf` with `pp` for pretty-print, which is what `bf` was actually toggling. I also removed the color output cause we really don't need to support that

rubicon_ml/cli.py
ryanSoley: they're in the definitions of each of the `click.options` - you can view the full help message with `rubicon_ml search --help`

stephenpardy (approved):

PR #394, opened 2023-10-12
json query bugfix

ryanSoley:
## What
  * deals with a bug in `jsonpath-ng` where non-homogenous values cause issues when querying with comparison operators (<, <=, >, >=)
    * e.g. all metrics have a value - if some have int values and some have str values, even queries over just the int ones will process the str ones as well and error as you can not compare an int and a str with <, >, etc.
  * adds an additional `json_numeric` that removes all entities with a non-numeric "value"
  * query `json_numeric` instead of `json` when < or > is in the query

## How to Test
  * `python -m pytest tests/unit/client/test_rubicon_json_client.py`

### DIFF
--- a/rubicon_ml/client/rubicon_json.py
+++ b/rubicon_ml/client/rubicon_json.py
@@ -1,8 +1,21 @@
+import copy
+import numbers
 from typing import Any, Dict, List, Optional, Type, Union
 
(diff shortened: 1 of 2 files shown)


shania-m (approved):
LGTM

Issue #387, opened 2023-09-22
can `Rubicon` extend from `Base`?


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

it seems like we've reimplemented nearly all the `Base` functionality on the `Rubicon` object at this point

**Describe the solution you'd like**

we should just inherit from the `Base` if possible

**Describe alternatives you've considered**

keep the duplicated code as-is

ryanSoley:
no - we will leave this as is

PR #396, opened 2023-11-14
Multiple backends refactor

stephenpardy:
---

## What
  * Simplifies the multiple backends setup so that users pass in multiple `Config` options.

## How to Test
  * Running existing unit tests

### DIFF
--- a/rubicon_ml/client/rubicon.py
+++ b/rubicon_ml/client/rubicon.py
@@ -100,9 +100,9 @@ class Rubicon:
     def _create_project_domain(
         self,
         name: str,
-        description: str,
-        github_url: str,
-        training_metadata: Union[List[Tuple], Tuple],
+        description: Optional[str],
+        github_url: Optional[str],
+        training_metadata: Optional[Union[List[Tuple], Tuple]],
     ):
         """Instantiates and returns a project domain object."""
         if self.config.is_auto_git_enabled and github_url is None:
@@ -121,7 +121,13 @@ class Rubicon:
         )
 
     @failsafe
-    def create_project(self, name, description=None, github_url=None, training_metadata=None):
+    def create_project(
+        self,
+        name: str,
+        description: Optional[str] = None,
+        github_url: Optional[str] = None,
+        training_metadata: Optional[Union[Tuple, List[Tuple]]] = None,
+    ) -> Project:
         """Create a project.
 
         Parameters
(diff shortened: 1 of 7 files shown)


rubicon_ml/client/rubicon.py
```diff
@@ -73,9 +100,9 @@ def _get_github_url(self):
...
         name: str,
-        description: str,
-        github_url: str,
-        training_metadata: Union[List[Tuple], Tuple],
+        description: Optional[str],
+        github_url: Optional[str],
+        training_metadata: Optional[Union[List[Tuple], Tuple]],
     ):
         """Instantiates and returns a project domain object."""
         if self.config.is_auto_git_enabled and github_url is None:
```
ryanSoley, line 108: this can be `self.is_auto_git_enabled` now, right?

ryanSoley (commented):
looking good so far!

rubicon_ml/client/rubicon.py
stephenpardy: Related to issue #387 - that method is in the `Base` class not here. Let me look.

rubicon_ml/client/rubicon.py
```diff
@@ -38,24 +38,51 @@ def __init__(
...
         **storage_options,
     ):
-        self.config = Config(persistence, root_dir, auto_git_enabled, **storage_options)
+        if composite_config is not None:
+            self.configs = [
+                Config(
+                    persistence=config["persistence"],
+                    root_dir=config["root_dir"],
+                    auto_git_enabled=auto_git_enabled,
+                )
```
ryanSoley, line 50: this also needs `**storage_options`

I think we should ultimately just make users pass a list of `Config`s if they want to use multiple backends - what do you think? we don't need to do it in this PR

stephenpardy: Good catch. Yes I was hoping to split this into two PRs. This one is mostly backwards compatible, and that change would require an API change.

stephenpardy: Also the fact that this didn't cause any unit tests to fail indicates a gap - let me address that before we merge here.

ryanSoley: > Yes I was hoping to split this into two PRs. This one is mostly backwards compatible, and that change would require an API change.

perfect, sounds good!

> Also the fact that this didn't cause any unit tests to fail indicates a gap - let me address that before we merge here.

thanks!

PR #409, opened 2024-02-08
adding comment creation support for domain layer & base repo

sonaalthaker:
---

## What
  * Comment creation support for domain layer & base repo

## How to Test
  * `pytest tests`

### DIFF
--- a/rubicon_ml/domain/mixin.py
+++ b/rubicon_ml/domain/mixin.py
@@ -25,3 +25,18 @@ class TagMixin:
             A list of string tags to remove from this domain model.
         """
         self.tags = list(set(self.tags).difference(set(tags)))
+
+
+class CommentMixin:
+    """Adds comment support to a domain model."""
+
+    def add_comments(self, comments: List[str]):
+        """
+        Add new comments to this model.
+
+        Parameters
+        ----------
+        comments : List[str]
+            A list of string comments to add to the domain model.
+        """
+        self.comments = list(set(self.comments).union(set(comments)))
--- a/rubicon_ml/domain/mixin.py
+++ b/rubicon_ml/domain/mixin.py
@@ -39,4 +39,4 @@ class CommentMixin:
         comments : List[str]
             A list of string comments to add to the domain model.
         """
-        self.comments = list(set(self.comments).union(set(comments)))
+        self.comments.extend(comments)
--- a/rubicon_ml/repository/base.py
+++ b/rubicon_ml/repository/base.py
@@ -1212,3 +1212,64 @@ class BaseRepository:
         sorted_tag_data = [json.loads(tag_data[p]) for _, p in sorted_tag_paths]
 
         return sorted_tag_data
+
+    # ---------- Comments ----------
+
+    def _get_comment_metadata_root(
+        self, project_name, experiment_id=None, entity_identifier=None, entity_type=None
+    ):
+        """Returns the directory to write comments to."""
+        get_metadata_root_lookup = {
+            "Artifact": self._get_artifact_metadata_root,
+            "Dataframe": self._get_dataframe_metadata_root,
+            "Experiment": self._get_experiment_metadata_root,
+            "Metric": self._get_metric_metadata_root,
+            "Feature": self._get_feature_metadata_root,
+            "Parameter": self._get_parameter_metadata_root,
+        }
+
+        try:
+            get_metadata_root = get_metadata_root_lookup[entity_type]
+        except KeyError:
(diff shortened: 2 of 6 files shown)


rubicon_ml/domain/mixin.py
```diff
@@ -25,3 +25,18 @@ def remove_tags(self, tags: List[str]):
...
+    def add_comments(self, comments: List[str]):
+        """
+        Add new comments to this model.
+
+        Parameters
+        ----------
+        comments : List[str]
+            A list of string comments to add to the domain model.
+        """
+        self.comments = list(set(self.comments).union(set(comments)))
```
ryanSoley, line 42: this was originally in tags because we didn't want duplicate tags. comments are pretty unstructured, so I don't think its necessary to guard against duplicates here. let's just do `self.comments.extend(comments)`

rubicon_ml/repository/base.py
```diff
@@ -1212,3 +1212,64 @@ def get_tags(self, project_name, experiment_id=None, entity_identifier=None, ent
         sorted_tag_data = [json.loads(tag_data[p]) for _, p in sorted_tag_paths]
 
         return sorted_tag_data
+
+    # ---------- Comments ----------
+
+    def _get_comment_metadata_root(
```
ryanSoley, line 1218: is there any difference between this and `_get_tag_metadata_root`? the metadata root is gonna be the same - you're just changing "tag" in the filename to "comment", which happens in the function you added below this one. if so, you can just have this function call `_get_tag_metadata_root`

rubicon_ml/repository/base.py
```diff
@@ -1212,3 +1212,64 @@ def get_tags(self, project_name, experiment_id=None, entity_identifier=None, ent
...
+            The ID of the experiment to apply the comments
+            `comments` to.
+        entity_identifier : str, optional
+            The ID or name of the entity to apply the comments
+            `comments` to.
+        entity_type : str, optional
+            The name of the entity's type as returned by
+            `entity_cls.__class__.__name__`.
+        """
+        comment_metadata_root = self._get_comment_metadata_root(
```
ryanSoley, line 1270: if my comment above is correct, you could just change this to `_get_tag_metadata_root` and leave a comment explaining that they're the same root path too. then just totally get rid of the above function. whichever is easier for you

tests/unit/repository/test_base_repo.py
```diff
@@ -980,3 +980,25 @@ def test_get_tags_with_no_results(memory_repository):
...
 
     assert tags == []
+
+
+def test_add_comments(memory_repository):
+    repository = memory_repository
+    experiment = _create_experiment(repository)
+    repository.add_comments(
+        experiment.project_name,
+        ["wow"],
```
ryanSoley, line 990: can we make this value a few words long? just wanna validate that its not splitting on spaces or anything like that

ryanSoley (commented):
looking great! just a few small questions

ryanSoley:
also not sure why the build isn't triggering. try rebasing off main and see if that fixes it

rubicon_ml/repository/base.py
sonaalthaker: Yeah, there's no real difference. I just thought it might be good form to add a comment specific one in case we wanted to differentiate between the two more in the future, but I think it's fine to remove now.

rubicon_ml/domain/mixin.py
sonaalthaker: I'll remove it, but just for the sake of curiosity - is there ever a situation we would actually want duplicates? Even if the check isn't as necessary, does having it actually hurt anything?

ryanSoley: I could imagine multiple users having a conversation and there being multiple "yes"s, "no"s. etc. but then again we'd probably want to make these more robust (like including authors) if we really wanted to support conversations like that

also, the list/union/list conversion is O(n) so it could become slow. but that'd require a lot of comments

rubicon_ml/repository/base.py
ryanSoley: yeah that definitely makes sense, and I could see the comments becoming a bit more complicated. if we just do something like 
```
def _get_comment_metadata_root(args):
    return self._get_tag_metadata_root(args)
```

we can have the best of both worlds - a new function available if it needs to change and reuse of the old one

ryanSoley:
looks like the linting failed. run `pre-commit install` in your local repo to make sure all the formatting checks run before you commit. it might error and complain about git hooks. if it does, just slack me

ryanSoley (approved):

PR #412, opened 2024-02-21
client layer comment support & comment removal support

sonaalthaker:
---

## What
  * client layer comment support & comment removal support

## How to Test
  * `pytest tests`

### DIFF
--- a/rubicon_ml/client/__init__.py
+++ b/rubicon_ml/client/__init__.py
@@ -1,12 +1,12 @@
 from rubicon_ml.client.base import Base  # noqa F401
 from rubicon_ml.client.config import Config
 
-from rubicon_ml.client.mixin import (
+from rubicon_ml.client.mixin import (  # noqa F401
     ArtifactMixin,
     DataframeMixin,
     TagMixin,
     CommentMixin,
-)  # noqa F401
+)
 
 from rubicon_ml.client.artifact import Artifact
 from rubicon_ml.client.dataframe import Dataframe
--- a/rubicon_ml/client/__init__.py
+++ b/rubicon_ml/client/__init__.py
@@ -1,7 +1,12 @@
 from rubicon_ml.client.base import Base  # noqa F401
 from rubicon_ml.client.config import Config
 
-from rubicon_ml.client.mixin import ArtifactMixin, DataframeMixin, TagMixin  # noqa F401
+from rubicon_ml.client.mixin import (
+    ArtifactMixin,
+    DataframeMixin,
+    TagMixin,
+    CommentMixin,
+)  # noqa F401
 
 from rubicon_ml.client.artifact import Artifact
 from rubicon_ml.client.dataframe import Dataframe
--- a/rubicon_ml/client/mixin.py
+++ b/rubicon_ml/client/mixin.py
@@ -604,3 +604,102 @@ class TagMixin:
                 return self._domain.tags
 
         self._raise_rubicon_exception(return_err)
+
+
+class CommentMixin:
+    """Adds comment support to a client object."""
+
+    _domain: DOMAIN_TYPES
+
+    def _get_taggable_identifiers(self):
+        project_name, experiment_id = self._parent._get_identifiers()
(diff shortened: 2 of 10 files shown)


setup.cfg
```diff
@@ -78,6 +78,7 @@ ignore =
 	W503
 	E203
 	E501
+	F401
```
sonaalthaker, line 81: Not sure if it's better to do this or specifically exclude `rubicon_ml/client/__init__.py`. There's an issue with the imports in this file falsely being marked as unused.

rubicon_ml/client/__init__.py
```diff
@@ -1,7 +1,12 @@
 from rubicon_ml.client.base import Base  # noqa F401
 from rubicon_ml.client.config import Config
 
-from rubicon_ml.client.mixin import ArtifactMixin, DataframeMixin, TagMixin  # noqa F401
+from rubicon_ml.client.mixin import (
+    ArtifactMixin,
+    DataframeMixin,
+    TagMixin,
+    CommentMixin,
+)  # noqa F401
```
ryanSoley, line 9: I think if you move the `# noqa F401` to line 4 after `import (` the linter will pick it up and ignore these lines without needing to update the global ignore list

rubicon_ml/client/mixin.py
```diff
@@ -604,3 +604,102 @@ def tags(self) -> List[str]:
...
 
         self._raise_rubicon_exception(return_err)
+
+
+class CommentMixin:
+    """Adds comment support to a client object."""
+
+    _domain: DOMAIN_TYPES
+
+    def _get_taggable_identifiers(self):
```
ryanSoley, line 614: can we update this name to reflect comments rather than tags

rubicon_ml/client/mixin.py
```diff
@@ -604,3 +604,102 @@ def tags(self) -> List[str]:
...
+        Parameters
+        ----------
+        comments : list of str
+             The comment values to remove.
+        """
+        project_name, experiment_id, entity_identifier = self._get_taggable_identifiers()
+
+        self._domain.remove_comments(comments)
+        for repo in self.repositories:
+            repo.remove_comments(
```
ryanSoley, line 669: there is no `remove_comments` on the repo, is there? I think we said we weren't going to do removals for comments. at least not for now

setup.cfg
ryanSoley: try removing this and doing what I mentioned above - if that doesn't work then this is cool

ryanSoley (commented):
looking good, although I think we can remove everything referencing comment removal

rubicon_ml/client/mixin.py
sonaalthaker: The way we implement `_update_tags` and `_update_comments` means that if we don't have the remove methods we end up adding two of each comment/tag. I figured adding `remove_comments` fixes that issue & and it doesn't hurt to have the option to remove comments in general even if it's not used much.

ryanSoley: I think it'd be better to update `_update_comments` and remove the references to deleting them. I'm not sure why it'd be adding multiples of each comment in its current state, so if that's occurring we should fix it without adding deletes

there is no `remove_comments` on the repository currently, so adding in deletes would add another chunk of work. I'd rather not expand the scope of this story at the moment

tests/unit/client/test_mixin_client.py
```diff
@@ -537,3 +542,12 @@ def raise_error():
     with pytest.raises(RubiconException) as e:
         experiment.tags()
     assert "all configured storage backends failed" in str(e)
+
+
+def test_add_comments(project_client):
```
ryanSoley, line 547: can we add client level tests for removing comments and the error message as well? like the last two tag tests here:
> https://github.com/capitalone/rubicon-ml/blob/main/tests/unit/client/test_mixin_client.py#L519

ryanSoley (approved):
nice job, thanks for all the changes!

PR #415, opened 2024-02-27
log `h2o` models as artifacts

ryanSoley:
## What
  * adds `log_h2o_model` to the artifact mixin, allowing projects and experiments to log `h2o` models as artifacts
    * `h2o` models cannot be pickled - they must be logged with `h2o.save_model`
    * there are other models that follow a similar pattern - we should see if this is more generalizable in the future

## How to Test
  * `python -m pytest`

### DIFF
--- a/rubicon_ml/client/artifact.py
+++ b/rubicon_ml/client/artifact.py
@@ -82,8 +82,8 @@ class Artifact(Base, TagMixin):
             * "h2o" to use `h2o.load_model` to load the data.
             Defaults to None.
         unpickle : bool, optional
-            Flag indicating whether artifact data must be
-            unpickled. Will be returned as bytes by default.
+            Flag indicating whether or not to unpickle artifact data.
+            `deserialize` takes precedence. Defaults to False.
         """
         project_name, experiment_id = self.parent._get_identifiers()
         return_err = None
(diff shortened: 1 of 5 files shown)


rubicon_ml/client/mixin.py
```diff
@@ -189,6 +190,51 @@ def log_conda_environment(self, artifact_name: Optional[str] = None) -> Artifact
...
+                export_cross_validation_predictions=export_cross_validation_predictions,
+                filename=artifact_name,
+                path=temp_dir_name,
+            )
+
+            artifact = self.log_artifact(
+                name=artifact_name,
+                data_path=model_data_path,
+                **log_artifact_kwargs,
+            )
```
ryanSoley, line 234: this pattern is probably reusable. xgboost's model saving function immediately comes to mind. we should look into expanding and generalizing this

rubicon_ml/client/artifact.py
```diff
@@ -66,15 +66,24 @@ def _get_data(self):
             self._raise_rubicon_exception(return_err)
 
     @failsafe
-    def get_data(self, unpickle: bool = False):
+    def get_data(
+        self,
+        deserialize: Optional[Literal["h2o"]] = None,
```
dianelee217, line 71: is there a reason why this is a literal instead of just a boolean?

shania-m (approved):
LGTM

PR #416, opened 2024-03-04
adding commentmixin to remaining entities

sonaalthaker:
---

## What
  * adding commentmixin to remaining entities

## How to Test
  * `pytest tests`

### DIFF
--- a/rubicon_ml/client/experiment.py
+++ b/rubicon_ml/client/experiment.py
@@ -57,8 +57,14 @@ class Experiment(Base, ArtifactMixin, DataframeMixin, TagMixin, CommentMixin):
 
     @failsafe
     def log_metric(
-        self, name, value, directionality="score", description=None, tags=[], comments=[]
-    ):
+        self,
+        name: str,
+        value: float,
+        directionality: str = "score",
+        description: str = None,
+        tags: list[str] = [],
+        comments: list[str] = [],
+    ) -> Metric:
         """Create a metric under the experiment.
 
         Parameters
@@ -176,7 +182,14 @@ class Experiment(Base, ArtifactMixin, DataframeMixin, TagMixin, CommentMixin):
             return [m for m in self.metrics() if m.id == id][0]
 
     @failsafe
-    def log_feature(self, name, description=None, importance=None, tags=[], comments=[]):
+    def log_feature(
+        self,
+        name: str,
+        description: str = None,
+        importance: float = None,
+        tags: list[str] = [],
+        comments: list[str] = [],
+    ) -> Feature:
         """Create a feature under the experiment.
 
         Parameters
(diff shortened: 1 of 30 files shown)


rubicon_ml/client/experiment.py
```diff
@@ -56,7 +56,9 @@ def _get_identifiers(self):
         return self.project.name, self.id
 
     @failsafe
-    def log_metric(self, name, value, directionality="score", description=None, tags=[]):
+    def log_metric(
+        self, name, value, directionality="score", description=None, tags=[], comments=[]
```
stephenpardy, line 60: Nit: Can we please add mypy type hints when we update the functions.

rubicon_ml/client/experiment.py
```diff
@@ -161,7 +176,7 @@ def metric(self, name=None, id=None):
             return [m for m in self.metrics() if m.id == id][0]
 
     @failsafe
-    def log_feature(self, name, description=None, importance=None, tags=[]):
+    def log_feature(self, name, description=None, importance=None, tags=[], comments=[]):
```
stephenpardy, line 179: Same as above.

tests/unit/client/test_dataframe_client.py
```diff
@@ -9,14 +9,18 @@ def test_properties(project_client):
     parent = project_client
 
     domain_dataframe = domain.Dataframe(
-        description="some description", tags=["x"], name="test title"
+        description="some description",
+        tags=["x"],
+        comments=["this is a comment"],
+        name="test title",
     )
```
ryanSoley, line 16: there's a `test_properties` in each of the client entities' unit tests - can we update them all like this?

looks like you've got dataframe and experiment - so we'll need artifact, feature, metric and parameter too

tests/unit/client/test_mixin_client.py
```diff
@@ -339,9 +339,15 @@ def test_log_dataframe(project_client, test_dataframe):
     project = project_client
     df = test_dataframe
     test_df_name = "test_df"
-    dataframe = DataframeMixin.log_dataframe(project, df, name=test_df_name, tags=["x"])
-    DataframeMixin.log_dataframe(project, df, name="secondary test df", tags=["x"])
+    dataframe = DataframeMixin.log_dataframe(
```
ryanSoley, line 342: can you test the same for the `ArtifactMixin` in the tests at the beginning of the file?

ryanSoley (commented):
looks great! just wanna see a few more tests. they should just be copies of the ones you already have

also, I tried logging a project with comments and it didn't work. the comments logged without error, but they weren't actually written to the correct location. when I tried `project.comments` after I got an error from `_get_parent_identifiers`. this is because projects don't have a parent and tags are not applied to projects - so we never encountered this issue with the `TagMixin`

we can either revert the project changes for this PR and revisit them later, or if you want to you can update `_get_parent_identifiers` in the `CommentMixin` to properly handle projects. up to you - lmk what you want to do

ryanSoley (approved):
nice work!

Issue #424, opened 2024-04-08
ExperimentsTable visualization fails when commit_hash==""


AdamAbate-6:
**Describe the bug**
The `ExperimentsTable` class has a method `load_experiment_data()`. This method populates `self.commit_hash `and `self.github_url` within the `if` block of line 281 if and only if the method's local variable `commit_hashes` is a set with one element. Taking a step back, `commit_hashes` is populated by iterating across the experiments and on line 247 doing `commit_hashes.add(experiment.commit_hash)`. _However_, line 247 is only reached `if experiment.commit_hash is not None`. So if `experiment.commit_hash` is an _empty string_ in every experiment, then `commit_hashes` will have only one element -- the empty string. That means the method will enter the `if` block of line 281.

Within that if block, if `self.experiments[0].project.github_url` is not None, nothing will error. But it is likely that if an experiment's commit_hash is an empty string, the user also did not specify any git info including the github_url. And if the github_url is None, then line 282 will error because you are indexing a None: `github_url_root = self.experiments[0].project.github_url[:-4]` 

**Steps/Code to reproduce bug**
Create experiments such that their commit_hash attributes are empty strings and their project.github_url attributes are None and verify via the following:
`[e.commit_hash for e in project.experiments()]`
`[e.project.github_url for e in project.experiments()]`

Then the following command will error:
`ExperimentsTable(experiments=project.experiments()).show()`

**Expected behavior**
If experiments' commit_hash values are empty strings, `ExperimentsTable` should behave just as if the commit_hash values were Nones, regardless of what any experiment's project.github_url attribute is set to.

**Additional context**
Let me know if any is needed.

ryanSoley:
thanks for the detailed bug report! we'll check it out asap

PR #426, opened 2024-04-09
viz bugfix

ryanSoley:
closes: #424 

---

## What
  * `self.is_auto_git_enabled` is a function, not a property - call it
  * make sure `commit_hash` is not empty string before attempting to visualize it

## How to Test
  * `python -m pytest tests/unit/viz`
  * run the experiment table viz notebook


shania-m (approved):
LGTM

PR #429, opened 2024-04-11
adds a `TagContainer` to handle key-value tags

### LINKED ISSUE #402: Use dictionary for Rubicon resource tagging
**Is your feature request related to a problem? Please describe**
When managing large amounts of Rubicon resources, users need some amount of database-like querying using tags.  
Querying resources would be easier if tagging used a key/value data structure.

**Describe the solution you'd like**
Implement tagging with a dictionary data structure for Rubicon resources (Experiments, Artifacts, etc). 

**Describe alternatives you've considered**
Today, users need to write their own custom code to search through strings. 
For example, an experiment is logged for each round of RFE using the pattern "RFE-iter-{iteration}".  

If we want the second round of RFE, we have to use:
rfe_tag = [tag for tag in experiment.tags if tag.startswith("RFE-iter-")][0]
rfe_tag[9:]

With key/value tags, we could access the rfe iteration with something like:
experiment.tags["rfe_iteration"] 

**Additional context**
Add any other context, code examples, or references to existing implementations about the feature request here.

ryanSoley:
closes #402 

---

## What
  * adds a `TagContainer` that extends from the native Python `list`
    * allows for indexing by string
      * when `tags` are indexed with a string, all tags with a `<key>:<value>` format will be considered and all `<value>`s corresponding to the string `<key>` will be returned
  * leverages this new functionality in `_get_experiments_from_tags`

## How to Test
  * `python -m pytest tests`
  * run the updated `tagging.ipynb` notebook

### DIFF
--- a/rubicon_ml/client/mixin.py
+++ b/rubicon_ml/client/mixin.py
@@ -687,7 +687,7 @@ class TagMixin:
             self._domain.remove_tags(tag.get("removed_tags", []))
 
     @property
-    def tags(self) -> List[str]:
+    def tags(self) -> TagContainer:
         """Get this client object's tags."""
         project_name, experiment_id, entity_identifier = self._get_taggable_identifiers()
         return_err = None
--- a/rubicon_ml/client/mixin.py
+++ b/rubicon_ml/client/mixin.py
@@ -16,7 +16,7 @@ import fsspec
 
 from rubicon_ml import client, domain
 from rubicon_ml.client.utils.exception_handling import failsafe
-from rubicon_ml.client.utils.tags import filter_children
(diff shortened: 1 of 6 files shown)


rubicon_ml/client/mixin.py
```diff
@@ -704,7 +704,7 @@ def tags(self) -> List[str]:
             else:
                 self._update_tags(tag_data)
 
-                return self._domain.tags
+                return TagContainer(self._domain.tags)
```
stephenpardy, line 707: Need to update the return type above?

apeterson7 (approved):
lgtm!

stephenpardy (approved):

dianelee217 (approved):

PR #437, opened 2024-05-23
programmatically select columns

ryanSoley:
## What
  * adds the ability to filter the parameters and metrics shown prior to rendering the experiment table
    * (rather than just being able to toggle them via the widget itself)

## How to Test
- [x] TODO: add new tests
- `python -m pytest tests/unit/viz/test_experiments_table.py`


stephenpardy (commented):
Can we please add some tests for this?

stephenpardy (approved):

Issue #444, opened 2024-06-14
publish visualizations via intake


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

we need to publish viz widgets in the catalogs before we can read them in #441 and #442

**Describe the solution you'd like**

`rubicon_ml.publish` should optionally take in a single viz object and publish it as an entry in the output catalog. the arguments for the entry should be all the arguments that are input into the viz object itself so we can recreate it later. the name can just be a placeholder for now

```yaml
sources:
  experiment_1b7b2bbf_b128_4eff_9eef_154f4da079e2:
    args:
      experiment_id: 1b7b2bbf-b128-4eff-9eef-154f4da079e2
      project_name: classifying penguins
      urlpath: ./rubicon-root
    driver: rubicon_ml_experiment
  experiment_cdd11d98_204f_4664_9110_6f73d06b9d15:
    args:
      experiment_id: cdd11d98-204f-4664-9110-6f73d06b9d15
      project_name: classifying penguins
      urlpath: ./rubicon-root
    driver: rubicon_ml_experiment
  # NEW SECTION STARTS HERE
  experiment_table:
    args:
      metric_names: ...
      ...
    driver: rubicon_ml_experiment_table
```

**Additional context**

in the future we may only need to take in viz objects. we can just get the experiments from `viz.experiments`

ryanSoley:
closed via #450

PR #449, opened 2024-06-20
Update publish.py

yashbhedaru:
closes: #444 

Updated publish.py to take in a visualization object is needed which passes relevant visualization parameters through to the catalog file.


rubicon_ml/intake_rubicon/publish.py
```diff
@@ -4,6 +4,8 @@
 
 def publish(
     experiments,
+    # visualization object passed, defaulted to None 
+    visualization_object=None,
```
stephenpardy, line 8: We have been working to expand type hinting throughout the library, would request that we add mypy type hints to any parameters we are adding.

rubicon_ml/intake_rubicon/publish.py
```diff
@@ -118,5 +122,24 @@ def _build_catalog(experiments):
 
         experiment_catalog_name = f"experiment_{experiment.id.replace('-', '_')}"
         catalog["sources"][experiment_catalog_name] = appended_experiment_catalog
+    
+    # create visualization entry to the catalog file
+    if visualization != None:
+        appended_visualization_catalog = {
+            "driver": "rubicon_ml_experiment_table",
+            "args": {
+                "experiments": visualization.experiments,
```
ryanSoley, line 131: we won't need the experiments themselves. we'll assume they're present as actual experiment sources in the catalog too, and later add some checks to validate that

ryanSoley (changes requested):
looking good! one small change request and the type hints like Stephen suggested

also, can you add a visualization object to this test?
> https://github.com/capitalone/rubicon-ml/blob/main/tests/unit/intake_rubicon/test_publish.py#L7

rubicon_ml/intake_rubicon/publish.py
```diff
@@ -118,5 +122,24 @@ def _build_catalog(experiments):
 
         experiment_catalog_name = f"experiment_{experiment.id.replace('-', '_')}"
         catalog["sources"][experiment_catalog_name] = appended_experiment_catalog
+    
+    # create visualization entry to the catalog file
+    if visualization != None:
```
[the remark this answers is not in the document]
ryanSoley: @yashbhedaru you can run `pre-commit install` to have the [style checks](https://capitalone.github.io/rubicon-ml/contributing.html#style) run on each commit

rubicon_ml/intake_rubicon/publish.py
yashvb: So would the code look something like this? " visualization_object: Optional[ExperimentsTable] = None" and I import the experiments table from the viz library into publish.py?

rubicon_ml/intake_rubicon/publish.py
yashvb: Okay I'll remove experiments from it.

rubicon_ml/intake_rubicon/publish.py
ryanSoley: yup! if you don't need the `ExperimentsTable` anywhere other than the type hint, you can guard the import behind a `TYPE_CHECKING` block
> https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

ryanSoley: that way, it'll only actually get imported when you're type checking. not at runtime where its not actually needed

ryanSoley:
closing in favor of #450

PR #450, opened 2024-06-21
Add visualization widgets to catalog through publish.py

yashvb:
**Description**
Issue: Currently we are trying to build an extension for serving visualizations through a Juptyer lab extension. To do this we want to include visualization information in the catalogs. 

Solution: For this issue, rubicon_ml.publish.py now optionally takes in a single viz object (ExperimentTable only at the moment) and publishs it as an entry in the output catalog. The arguments for the entry are all the arguments that are input into the viz object itself so it can be recreated later. 

**Changes**
1) The publish() function now takes in a visualization object which is defaulted to None, added type hinting using mypy (only for ExperimentsTable object)
2) Added type checking for the ExperimentsTable type such that it only gets imported during type checking.
3) Incorporated a new visualization functionality to update_catalog() such that it will build a new catalog with the new visualization object if there is one passed, otherwise just the experiments. 
4) Updated build_catalog() such that a new visualization entry is appended to the catalog file in yaml format with the parameters inputted into the visualization object. 

**Related Issues**
[#444](https://github.com/capitalone/rubicon-ml/issues/444) -> Original Issue discussing changes needed and background
[#449 ](https://github.com/capitalone/rubicon-ml/pull/449) -> Original Pull request with initial feedback on 1st commit for changes

### DIFF
--- a/rubicon_ml/intake_rubicon/publish.py
+++ b/rubicon_ml/intake_rubicon/publish.py
@@ -1,9 +1,16 @@
+from typing import TYPE_CHECKING, Optional
+
 import fsspec
 import yaml
+
+if TYPE_CHECKING:
+    from rubicon_ml.viz.experiments_table import ExperimentsTable
 
 
 def publish(
     experiments,
+    # visualization object passed, defaulted to None
+    visualization_object: Optional["ExperimentsTable"] = None,
     output_filepath=None,
     base_catalog_filepath=None,
 ):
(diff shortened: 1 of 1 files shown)


yashvb:
Additionally, updated https://github.com/capitalone/rubicon-ml/blob/main/tests/unit/intake_rubicon/test_publish.py#L7 to have the ability to pass in a visualization object for testing our new publish.py. It is defaulted to None and can only handle the ExperimentsTable Type at the moment.

ryanSoley (approved):
nice job!

rubicon_ml/intake_rubicon/publish.py
```diff
@@ -1,9 +1,16 @@
...
 import fsspec
 import yaml
 
+if TYPE_CHECKING:
+    from rubicon_ml.viz.experiments_table import ExperimentsTable
+
 
 def publish(
     experiments,
+    # visualization object passed, defaulted to None
```
stephenpardy, line 12: I appreciate this comment, but types are self documenting and this is not needed going forward.

stephenpardy (approved):

Issue #442, opened 2024-06-13
create intake source for experiment table


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

we would like to be able to share [visualizations](https://capitalone.github.io/rubicon-ml/visualizations.html) alongside the projects and experiments in the intake catalogs. for example, `catalog.read()` returns projects and experiments. we'd like to ultimately have something like `catalog.serve()` that mimics the viz module's [`serve` functionality](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/viz/base.py#L61) to launch a dashboard with the experiments returned by `catalog.read()`

in #441 we created a base source for visualizations. we will need a concrete implementation for each visualization, like we have [for experiments](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/intake_rubicon/experiment.py#L6). this issue will track the creation of a source for [the experiments table](https://capitalone.github.io/rubicon-ml/visualizations/experiments-table.html)

**Describe the solution you'd like**

add a new `ExperimentsTableDataSource` that extends `VizDataSourceMixin`. the `__init__` function should take in any additional arguments specific to the `ExperimentsTable`. `_get_schema` should create a visualization and set it as the `_visualization` attribute, as well as update the metadata with any required attributes

**Additional context**

at this point, the visualization stored in `_visualization` does not need to have any experiments in it - we'll get to that later

ryanSoley:
closed w/ #456

Issue #441, opened 2024-06-13
create base intake source for viz objects


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

we would like to be able to share [visualizations](https://capitalone.github.io/rubicon-ml/visualizations.html) alongside the projects and experiments in the intake catalogs. for example, `catalog.read()` returns projects and experiments. we'd like to ultimately have something like `catalog.serve()` that mimics the viz module's [`serve` functionality](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/viz/base.py#L61) to launch a dashboard with the experiments returned by `catalog.read()`

to get started, we will need a base intake source for rubicon-ml visualizations. this issue will track the creation of a base visualization data source similar to the existing one for projects and experiments
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/intake_rubicon/base.py

**Describe the solution you'd like**

add a new `VizDataSourceMixin` that extends `intake.source.base.DataSource`. the `__init__` function should take in any arguments that will be required for all visualizations. [the `VizBase` itself](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/viz/base.py#L19) doesn't take anything other than a title, so it might just be that for this base class

the new class' `_get_schema` should be implemented the same as the existing `VizDataSourceMixin` - we will just leverage the extra metadata to hold the info we need. `read` and `_close` should set and clean up a `_visualization` attribute, similar to how the current class operates on `_rubicon_object`. the value of `_visualization` itself will be set in each concrete subclass of this base class

ryanSoley:
closed via #451

PR #456, opened 2024-06-27
Created an Intake Source for Experiment Table

yashvb:
**Description**
Issue: Currently we are trying to build an extension for serving visualizations through a Juptyer lab extension. To do this we want to include visualization information in the catalogs. Initially, we created a base visualization data source similar to the existing one for projects and experiments. Now we need to create a concrete implementation for Experiments Table. 
[base.py](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/intake_rubicon/base.py)

Solution: For this issue, we added a new ExperimentsTableDataSource that is going to extend VizDataSourceMixin. We will read in the catalog data and metadata from the passed arguments . Then `_get_schema` creates a experiments Table visualization and sets it as the visualization object attribute. 

**Changes**

1) Created a new file called `viz.py` within the intake_rubicon folder. This file holds the code for the new `ExperimentsTableDataSource`. 
2) Updated the `__init__ ` function within the intake_rubicon folder to now be able to include ExperimentsTableDataSource for module importing purposes. 
3) Updated   `rubicon_ml/viz/experiments_table.py` to import the publish function straight from the intake_rubicon folder as there was a circular import issue. 
4) Updated the `setup.cfg` file to include the ExperimentsTableDataSource as an Intake driver. 

**Related Issues**
[#442](https://github.com/capitalone/rubicon-ml/issues/442) -> Original Issue discussing changes needed and background
[#441](https://github.com/capitalone/rubicon-ml/issues/441) -> An extension of this issue.

### DIFF
--- a/rubicon_ml/intake_rubicon/__init__.py
+++ b/rubicon_ml/intake_rubicon/__init__.py
@@ -1,5 +1,6 @@
 import intake  # noqa F401
 
 from rubicon_ml.intake_rubicon.experiment import ExperimentSource
+from rubicon_ml.intake_rubicon.viz import ExperimentsTableDataSource
 
-__all__ = ["ExperimentSource"]
+__all__ = ["ExperimentSource", "ExperimentsTableDataSource"]
--- a/rubicon_ml/intake_rubicon/viz.py
+++ b/rubicon_ml/intake_rubicon/viz.py
@@ -1,0 +1,22 @@
+from rubicon_ml import __version__
+from rubicon_ml.intake_rubicon.base import VizDataSourceMixin
+from rubicon_ml.viz import ExperimentsTable
+
+
+class ExperimentsTableDataSource(VizDataSourceMixin):
+    """An Intake data source for reading `rubicon` Experiment Table visualizations."""
+
+    version = __version__
+
+    container = "python"
+    name = "rubicon_ml_experiments_table"
+
+    def __init__(self, metadata=None, **catalog_data):
+        self._catalog_data = catalog_data or {}
+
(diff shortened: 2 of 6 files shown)


ryanSoley (commented):
looks great! let's start adding some tests here before we merge this since we're in a more complete state now

first, let's actually test the values in the catalog in [test_publish.py](https://github.com/yashvb/rubicon-ml-test/blob/main/tests/unit/intake_rubicon/test_publish.py). instead of having `visualization_object` as an input with a `None` value, we'll just need to actually create one in the test function itself

I didn't realize when we originally merged those test changes, but inputs to tests are called [fixtures](https://github.com/yashvb/rubicon-ml-test/blob/main/tests/fixtures.py) and I'm surprised it didn't error because we don't have a fixture called `visualization_object` - it must be because you gave it a default. anyways, lets just add an assertion that `catalog["sources"]["experiment_table"]` exists when a viz object is passed in

then, we can add a test for the source itself. it should look similar to `test_source` in this experiment source test
> https://github.com/yashvb/rubicon-ml-test/blob/main/tests/unit/intake_rubicon/test_experiment.py#L12

we can add a new test file `tests/unit/intake_rubicon/test_viz.py` to contain the new `test_experiments_table_source` for reading visualizations from catalogs

ryanSoley:
also idk what's up with the mend failures. I don't think its anything on our end. it should be fine on the next commit

ryanSoley:
mend seems to be broken - you can just ignore it for now

ryanSoley (approved):
nice work!

Issue #458, opened 2024-06-28
Allow users to save models in ONNX format


stephenpardy:
**Is your feature request related to a problem? Please describe**
Currently Rubicon defaults to saving models in pickle. Pickles are known to be harmful in a number of cases. They are also not cross platform.

**Describe the solution you'd like**
Allow users to save models using ONNX. 

**Describe alternatives you've considered**
Handle custom formats like XGBoost's json format - this is handy, but would require custom handling for each model format. 

**Additional context**
https://onnx.ai/sklearn-onnx/auto_tutorial/plot_gexternal_xgboost.html

ryanSoley:
we should be able to do this at the client level similarly to the H2O artifact enhancements in #415

stephenpardy:
Some other comments:
- For ONNX, we would likely require users to pass in a function that does the conversion. See e.g. https://github.com/onnx/onnxmltools. Unless we write/find a utility that automatically grabs the correct converter for a given model type.

PR #459, opened 2024-06-28
Work to deprecate pickle flag

stephenpardy:
Starts work toward: #458 
---

## What
  * Moves the pickle loading functionality into the `deserialize` string. This is the first step of adding functionality for other model types. 
  * Followups include: xgboost support, ONNX support.

## How to Test
  * Run new and existing unit tests.

### DIFF
--- a/rubicon_ml/client/artifact.py
+++ b/rubicon_ml/client/artifact.py
@@ -51,7 +51,7 @@ class Artifact(Base, TagMixin, CommentMixin):
     def _get_data(self):
         """Loads the data associated with this artifact."""
         project_name, experiment_id = self.parent._get_identifiers()
-        return_err = RuntimeError("Data was not loaded properly.")
+        return_err = None
 
         self._data = None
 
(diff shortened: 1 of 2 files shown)


rubicon_ml/client/artifact.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, domain: ArtifactDomain, parent: Project):
     def _get_data(self):
         """Loads the data associated with this artifact."""
         project_name, experiment_id = self.parent._get_identifiers()
-        return_err = None
+        return_err = RuntimeError("Data was not loaded properly.")
```
ryanSoley, line 54: `return_err` is to re-raise errors from the repository layer - if there is an error this will always be overwritten in the `except` block below

rubicon_ml/client/artifact.py
```diff
@@ -83,13 +83,24 @@ def get_data(
...
+            * "pickle" to use pickles to load the data.
             Defaults to None.
         unpickle : bool, optional
             Flag indicating whether or not to unpickle artifact data.
             `deserialize` takes precedence. Defaults to False.
+            **Deprecated**: Please use `deserialize="pickle"` in the future.
         """
         project_name, experiment_id = self.parent._get_identifiers()
-        return_err = None
+        return_err = RuntimeError(
```
ryanSoley, line 94: same as above. this should probably just be its own check and `ValueError`

ryanSoley (commented):
thanks! left some comments around the error handling

ryanSoley (approved):

Issue #433, opened 2024-04-17
Support logging polars dataframes


stephenpardy:
**Is your enhancement request related to a problem? Please describe**
Polars is becoming a popular dataframe library due to its speed compared to pandas. Currently the dataframe logging of rubicon will not support polars due to the unique API.

**Describe the solution you'd like**
Add support for Polars in the `log_dataframe` method. We should detect polars and use the https://docs.pola.rs/py-polars/html/reference/api/polars.DataFrame.write_parquet.html method there.

**Describe alternatives you've considered**
Business as usual. 

**Additional context**
N/A

PR #460, opened 2024-07-01
Polars writing

stephenpardy:
closes: #433 

---

## What
  * Adds support for logging polars dataframes

## How to Test
  * Run the unit tests

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -39,6 +39,7 @@ dependencies:
   - pytest
   - pytest-cov
   - xgboost
+  - polars<1.0
   
   # for versioning
   - versioneer
--- a/rubicon_ml/client/mixin.py
+++ b/rubicon_ml/client/mixin.py
@@ -23,6 +23,7 @@ from rubicon_ml.exceptions import RubiconException
 if TYPE_CHECKING:
     import dask.dataframe as dd
     import pandas as pd
+    from polars import DataFrame as polars_DataFrame
 
     from rubicon_ml.client import Artifact, Dataframe
     from rubicon_ml.domain import DOMAIN_TYPES
@@ -455,7 +456,7 @@ class DataframeMixin:
     @failsafe
     def log_dataframe(
         self,
-        df: Union[pd.DataFrame, dd.DataFrame],
+        df: Union[pd.DataFrame, dd.DataFrame, polars_DataFrame],
         description: Optional[str] = None,
         name: Optional[str] = None,
         tags: Optional[List[str]] = None,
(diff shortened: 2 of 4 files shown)


environment.yml
```diff
@@ -39,6 +39,7 @@ dependencies:
   - pytest
   - pytest-cov
   - xgboost
+  - polars<1.0
```
ryanSoley, line 42: any particular reason for the pin?

rubicon_ml/client/mixin.py
```diff
@@ -23,6 +23,7 @@
 if TYPE_CHECKING:
     import dask.dataframe as dd
     import pandas as pd
+    from polars import DataFrame as polars_DataFrame
```
ryanSoley, line 26: nit: can we do the import consistently? either like this or with `import polars as pl` like the rest of the files

rubicon_ml/repository/base.py
```diff
@@ -602,9 +608,14 @@ def _persist_dataframe(self, df, path):
...
-        df.to_parquet(path, engine="pyarrow")
+        if hasattr(df, "write_parquet"):
+            # handle Polars
+            df.write_parquet(path)
+        else:
+            # Dask or pandas
+            df.to_parquet(path, engine="pyarrow")
 
-    def _read_dataframe(self, path, df_type="pandas"):
+    def _read_dataframe(self, path, df_type: Literal["pandas", "dask", "polars"] = "pandas"):
```
ryanSoley, line 618: this function needs to be updated to read polars still, right?

tests/unit/repository/test_base_repo.py
```diff
@@ -488,6 +523,23 @@ def test_create_dask_dataframe(memory_repository):
...
+
+    open_file = repository.filesystem.open(dataframe_metadata_path)
+    with open_file as f:
+        dataframe_json = json.load(f)
+
+    assert repository.filesystem.exists(dataframe_data_path)
+    assert dataframe.id == dataframe_json["id"]
+
+
 def test_get_pandas_dataframe(memory_repository):
```
ryanSoley, line 543: we need a getter test similar to this one for polars

environment.yml
stephenpardy: New version with some breaking changes just released. https://docs.pola.rs/releases/upgrade/1/ Would like to wait to double check before allowing this new version.

rubicon_ml/repository/base.py
```diff
@@ -709,7 +737,13 @@ def get_dataframes_metadata(
 
         return self._load_metadata_files(dataframe_metadata_root, domain.Dataframe)
 
-    def get_dataframe_data(self, project_name, dataframe_id, experiment_id=None, df_type="pandas"):
+    def get_dataframe_data(
+        self,
+        project_name: str,
+        dataframe_id: str,
+        experiment_id: Optional[str] = None,
+        df_type: Literal["pandas", "dask", "polars"] = "pandas",
```
ryanSoley, line 745: we can probably just infer this and store it in the metadata so users don't have to keep specifying

ryanSoley (approved):
looks great! thanks!

PR #461, opened 2024-07-01
Add XGBoost JSON Reader/writer

stephenpardy:
Related to #458 

---

## What
  * Adds ability to save and load models in xgboost's JSON format for better backwards compatibility

## How to Test
  * Run unit tests

### DIFF
--- a/rubicon_ml/client/mixin.py
+++ b/rubicon_ml/client/mixin.py
@@ -282,6 +282,42 @@ class ArtifactMixin:
         return artifact
 
     @failsafe
+    def log_xgboost_model(self, xgboost_model, artifact_name: Optional[str] = None, **log_artifact_kwargs: Any) -> Artifact:
+        """Log an XGBoost model as a JSON file to this client object.
+
+        Parameters
+        ----------
+        xgboost_model
+            An xgboost model object.
+        artifact_name : str, optional
+            The name of the artifact (the exported XGBoost model).
+        log_artifact_kwargs : Any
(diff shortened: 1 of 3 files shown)


ryanSoley:
a rebase should handle the test failures now

rubicon_ml/client/mixin.py
```diff
@@ -282,6 +283,49 @@ def log_h2o_model(
...
+    @failsafe
+    def log_xgboost_model(
+        self,
+        xgboost_model: "xgb.Booster",
+        artifact_name: Optional[str] = None,
+        **log_artifact_kwargs: Any,
+    ) -> Artifact:
+        """Log an XGBoost model as a JSON file to this client object.
+
+        Please note that we do not currently support logging directly from the SKLearn interface.
```
ryanSoley, line 295: what do you mean by this?

ryanSoley (approved):
looks good! just one question

ryanSoley:
closing and re-opening to try to get the license check to work

rubicon_ml/client/mixin.py
stephenpardy: If users use the `XGBRegressor` or `XGBClassifier` classes that implement sklearn interface for xgboost this will not work. They would have to use `model.get_booster()` first and log that. An obvious addition is to check for that ourselves and then call `model.get_booster()` if relevant. 

I can add that in this PR if you want.

ryanSoley: ahh gotcha, nah this is fine for now

PR #462, opened 2024-07-12
allow additional AWS CLI config via `sync`

ryanSoley:
## What
  * allows users to directly specify a AWS profile and credential location for the sync operation

## How to Test
  * `python -m pytest tests/unit/client/test_rubicon_client.py -k test_sync`

### DIFF
--- a/rubicon_ml/client/rubicon.py
+++ b/rubicon_ml/client/rubicon.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 import warnings
 from typing import Any, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


rubicon_ml/client/rubicon.py
```diff
@@ -298,7 +299,7 @@ def projects(self):
             raise return_err
 
     @failsafe
-    def sync(self, project_name, s3_root_dir):
+    def sync(self, project_name, s3_root_dir, aws_profile=None, aws_shared_credentials_file=None):
```
stephenpardy, line 302: Can we add type hints when we update this?

thebrianbn (approved):
LGTM

stephenpardy (approved):

PR #464, opened 2024-07-18
metric correlation plot intake source added

austinhk:
closes: #455 

---
Description
Issue: Currently we are trying to build an extension for serving visualizations through a Juptyer lab extension. To do this we want to include visualization information in the catalogs. Initially, we created a base visualization data source similar to the existing one for projects and experiments. Now we need to create a concrete implementation for Metric Correlation Plot.
[base.py](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/intake_rubicon/base.py)

Solution: For this issue, we added a new MetricCorrelationPlotDataSource that is going to extend VizDataSourceMixin. We will read in the catalog data and metadata from the passed arguments . Then _get_schema creates a metric correlation plot visualization and sets it as the visualization object attribute.

Changes

Modified file viz.py within the intake_rubicon folder. This file holds the code for the new MetricCorrelationPlotDataSource.
Updated the __init__  function within the intake_rubicon folder to now be able to include MetricCorrelationPlotDataSource for module importing purposes.
Updated the setup.cfg file to include the MetricCorrelationPlotDataSource as an Intake driver.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -45,7 +45,7 @@ dependencies:
   - versioneer
 
   # for packaging
-  - setuptools
+  - setuptools<71.0.0
   - wheel
 
   # for edgetest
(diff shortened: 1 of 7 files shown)


ryanSoley (approved):
great work! everything's running properly locally

**NOTE TO SELF** don't leave setuptools pinned for too long

PR #465, opened 2024-07-18
Dataframe plot

jeh362:
closes: #60

---
**Description**
Issue: Currently we are trying to build an extension for serving visualizations through a Juptyer lab extension. To do this we want to include visualization information in the catalogs. Initially, we created a base visualization data source similar to the existing one for projects and experiments. Now we need to create a concrete implementation for Dataframe Plots.

Solution: For this issue, we added a new `DataframePlotDataSource` that is going to extend `VizDataSourceMixin`. We will read in the catalog data and metadata from the passed arguments . Then _get_schema creates a Dataframe Plot visualization and sets it as the visualization object attribute.

**Changes**

1. Updated `viz.py` within the intake_rubicon folder. This file holds the code for the new `DataframePlotDataSource`.
2. Updated the `__init__ ` function within the intake_rubicon folder to now be able to include `DataframePlotDataSource` for module importing purposes.
3. Updated rubicon_ml/viz/publish.py to take in a `DataframePlot` as a visualization object
4. Updated the `setup.cfg` file to include the `DataframePlotDataSource` as an Intake driver.
5. Updated `test_viz.py` to include a `DataframePlotDataSource` test method as well.

### DIFF
--- a/rubicon_ml/intake_rubicon/viz.py
+++ b/rubicon_ml/intake_rubicon/viz.py
@@ -1,6 +1,8 @@
 from rubicon_ml import __version__
 from rubicon_ml.intake_rubicon.base import VizDataSourceMixin
 from rubicon_ml.viz import MetricCorrelationPlot
+from rubicon_ml.viz import DataframePlot
+from rubicon_ml.viz import ExperimentsTable
 
 
 class ExperimentsTableDataSource(VizDataSourceMixin):
@@ -18,7 +20,6 @@ class ExperimentsTableDataSource(VizDataSourceMixin):
 
     def _get_schema(self):
         """Creates an Experiments Table visualization and sets it as the visualization object attribute"""
-        from rubicon_ml.viz import ExperimentsTable
 
         self._visualization_object = ExperimentsTable(**self._catalog_data)
 
@@ -27,7 +28,6 @@ class ExperimentsTableDataSource(VizDataSourceMixin):
 
 class MetricCorrelationPlotDataSource(VizDataSourceMixin):
     """An Intake data source for reading `rubicon` Metric Correlation Plot visualizations."""
-
     version = __version__
 
     container = "python"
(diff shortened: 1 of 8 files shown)


rubicon_ml/intake_rubicon/viz.py
```diff
@@ -18,7 +21,6 @@ def __init__(self, metadata=None, **catalog_data):
 
     def _get_schema(self):
         """Creates an Experiments Table visualization and sets it as the visualization object attribute"""
-        from rubicon_ml.viz import ExperimentsTable
```
ryanSoley, line 21: can you move this import back into the `ExperimentsTableDataSource._get_schema` function?

can you also move the `from rubicon_ml.viz import DataframePlot` into `DataframePlotDataSource._get_schema` and `from rubicon_ml.viz import MetricCorrelationPlot` into `MetricCorrelationPlotDataSource._get_schema` (missed that when I reviewed @austinhk 's PR)

if we import them at the top level, it makes the viz dependencies required for the core package import, which we don't want. when we move them inside the relevant `_get_schema`, they'll only be imported when the visualizations are loaded from a catalog

ryanSoley:
one last thing - can we add a publish test like how `tests/unit/intake_rubicon/test_publish.py` was updated in #464

tests/unit/intake_rubicon/test_viz.py
```diff
@@ -58,3 +59,38 @@ def test_metric_correlation_plot_source():
...
+    )
+    assert source is not None
+
+    source.discover()
+
+    visualization = source.read()
+
+    assert visualization is not None
+    assert visualization.experiments == catalog_data_sample["dataframe_name"]
+    assert visualization.experiments == catalog_data_sample["experiments"]
```
ryanSoley, line 90:
```suggestion
    assert visualization.dataframe_name == catalog_data_sample["dataframe_name"]
```

first one needs to be `dataframe_name` and the second needs to be removed. the experiments on the viz object will be actual rubicon experiments and the experiments in the catalog are just ids. they won't match

ryanSoley (approved):
looks great!

PR #467, opened 2024-07-22
Adding MetricListComparisonDataSource

jeh362:
closes: #62

---
**Description**
Issue: Currently we are trying to build an extension for serving visualizations through a Juptyer lab extension. To do this we want to include visualization information in the catalogs. Initially, we created a base visualization data source similar to the existing one for projects and experiments. Now we need to create a concrete implementation for Metric List Comparison visualizations.

Solution: For this issue, we added a new `MetricListComparisonDataSource` that is going to extend `VizDataSourceMixin`. We will read in the catalog data and metadata from the passed arguments . Then `_get_schema` creates a Metric List Comparison visualization and sets it as the `visualization_object` attribute.

**Changes**

Updated `viz.py` within the intake_rubicon folder. This file holds the code for the new `MetricListComparisonDataSource`.
Updated the `__init__`  function within the intake_rubicon folder to now be able to include `MetricListComparisonDataSource` for module importing purposes.
Updated `rubicon_ml/viz/publish.py` to take in a `MetricListComparison` as a visualization object
Updated the `setup.cfg` file to include the `MetricListComparisonDataSource` as an Intake driver.
Updated` test_viz.py `to include a `MetricListComparisonDataSource` test method as well.
Updated `test_publish.py` to include a `MetricListsComparison` visualization object as well.
Updated `metric_lists_comparison.py` to fix certain typing errors.

### DIFF
--- a/rubicon_ml/intake_rubicon/publish.py
+++ b/rubicon_ml/intake_rubicon/publish.py
@@ -104,6 +104,7 @@ def _update_catalog(
 
 def _build_catalog(experiments, visualization):
     from rubicon_ml.viz.experiments_table import ExperimentsTable
+    from rubicon_ml.viz import DataframePlot
     from rubicon_ml.viz.metric_correlation_plot import MetricCorrelationPlot
 
     """Helper function to build catalog dictionary from given experiments.
--- a/rubicon_ml/intake_rubicon/publish.py
+++ b/rubicon_ml/intake_rubicon/publish.py
@@ -163,14 +163,14 @@ def _build_catalog(experiments, visualization):
             catalog["sources"]["metric_correlation_plot"] = appended_visualization_catalog
 
         # vizualization is an DataframePlot
-        if isinstance(visualization,DataframePlot):
+        if isinstance(visualization, DataframePlot):
             appended_visualization_catalog = {
                 "driver": "rubicon_ml_dataframe_plot",
                 "args": {
                     "dataframe_name": visualization.dataframe_name,
                     "x": visualization.x,
                     "y": visualization.y,
-                }
+                },
             }
 
             # append visualization object to end of catalog file
--- a/rubicon_ml/intake_rubicon/publish.py
+++ b/rubicon_ml/intake_rubicon/publish.py
@@ -107,7 +107,6 @@ def _update_catalog(
 
 def _build_catalog(experiments, visualization):
     from rubicon_ml.viz import DataframePlot
-    from rubicon_ml.viz import DataframePlot
     from rubicon_ml.viz.experiments_table import ExperimentsTable
     from rubicon_ml.viz.metric_correlation_plot import MetricCorrelationPlot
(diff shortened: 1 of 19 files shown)


rubicon_ml/intake_rubicon/publish.py
```diff
@@ -181,4 +183,20 @@ def _build_catalog(experiments, visualization):
 
         # append visualization object to end of catalog file
 
+        # vizualization is an MetricListsComparison
+        if isinstance(visualization, MetricListsComparison):
+            appended_visualization_catalog = {
+                "driver": "rubicon_ml_metric_list",
+                "args": {
+                    "column_names": visualization.column_names,
+                    "experiments": visualization.experiments,
```
ryanSoley, line 192: we don't want to log the experiments here. they'll be handled by the experiment source in the catalog like with the other viz sources

rubicon_ml/viz/metric_lists_comparison.py
```diff
@@ -78,7 +78,7 @@ def load_experiment_data(self):
 
         for experiment in self.experiments:
             for metric in experiment.metrics():
-                if isinstance(metric.value, list):
+                if isinstance(metric.value, float):
```
ryanSoley, line 81: we shouldn't be changing anything in this file - this needs to stay a list. you'll need to log list-type metrics to test if that's why this was changed

rubicon_ml/viz/metric_lists_comparison.py
```diff
@@ -154,7 +154,7 @@ def update_metric_heatmap(*args):
 
             data_array = np.array(heatmap_data)
             numerator = data_array - data_array.min(axis=0)
-            denominator = data_array.max(axis=0) - data_array.min(axis=0)
+            denominator = np.array(data_array.max(axis=0) - data_array.min(axis=0))
```
ryanSoley, line 157: same as above unless there's some specific reason this needs to be changed

ryanSoley (changes requested):
almost there - needs a few small changes

rubicon_ml/viz/metric_lists_comparison.py
jeh362: I was running into a couple errors before that this change fixed but I tried it again and it seemed to work now without this change.

rubicon_ml/intake_rubicon/publish.py
```diff
@@ -165,7 +167,7 @@ def _build_catalog(experiments, visualization):
             }
             catalog["sources"]["metric_correlation_plot"] = appended_visualization_catalog
 
-        # vizualization is an DataframePlot
+        # vizualization is an DataframePlot   tests/unit/intake_rubicon/test_publish.py
```
ryanSoley, line 170:
```suggestion
        # vizualization is a DataframePlot
```

ryanSoley (approved):
thanks! looks great!

Issue #468, opened 2024-07-25
`DataframePlot` - don't error if not all experiments have a dataframe


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

when sending a list of experiments to a `DataframePlot`, the viz will error if each experiment does not have a dataframe named `dataframe_name`

**Describe the solution you'd like**

instead of erroring, we should visualize the dataframes that are present and warn about the experiments that were not visualized:
* in [this loop](https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/viz/dataframe_plot.py#L93) in `load_experiment_data`, `dataframe = experiment.dataframe(name=self.dataframe_name)` should not immediately error if there is no dataframe for the current experiment
  * instead skip it, raise a warning about the experiment not having the dataframe, and move on to the next iteration
  * if `self.data_df` is still None at the end of the loop, then raise an error that no dataframe with name `dataframe_name` has been found

PR #471, opened 2024-07-25
DataframePlot - don't error if not all experiments have a dataframe 

jeh362:
closes: #468

---

**Issue**
When sending a list of experiments to a DataframePlot, the viz will error if each experiment does not have a dataframe with the name dataframe_name logged to it.

**Changes**
Updated `dataframe_plot.py` within the viz folder to display visualizations for available dataframes and skip experiments with no dataframes logged to them. If none of the experiments have dataframes, then an error is thrown.

### DIFF
--- a/rubicon_ml/viz/dataframe_plot.py
+++ b/rubicon_ml/viz/dataframe_plot.py
@@ -92,8 +92,10 @@ class DataframePlot(VizBase):
         count = 0
 
         for experiment in self.experiments:
-            if (len(experiment.dataframes())==0) :
-                print(f"WARNING: Experiment {experiment.name} does not have any dataframes logged to it. ")
+            if len(experiment.dataframes()) == 0:
+                print(
+                    f"WARNING: Experiment {experiment.name} does not have any dataframes logged to it. "
+                )
                 continue
             count += 1
             dataframe = experiment.dataframe(name=self.dataframe_name)
@@ -121,7 +123,7 @@ class DataframePlot(VizBase):
                 len(self.experiments),
             )
 
-        if (count == 0):
+        if count == 0:
             raise Exception(f"No dataframe with name {self.dataframe_name} found!")
 
     def register_callbacks(self, link_experiment_table=False):
--- a/rubicon_ml/viz/dataframe_plot.py
+++ b/rubicon_ml/viz/dataframe_plot.py
@@ -125,7 +125,7 @@ class DataframePlot(VizBase):
                 len(self.experiments),
             )
         if self.data_df is None:
-            raise Exception(f"No dataframe with name {self.dataframe_name} found!")
+            raise RubiconException(f"No dataframe with name {self.dataframe_name} found!")
 
     def register_callbacks(self, link_experiment_table=False):
         outputs = [
--- a/rubicon_ml/viz/dataframe_plot.py
+++ b/rubicon_ml/viz/dataframe_plot.py
@@ -95,7 +95,7 @@ class DataframePlot(VizBase):
         for experiment in self.experiments:
             try:
                 dataframe = experiment.dataframe(name=self.dataframe_name)
-            except:
+            except Exception:
                 warnings.warn(
                     f"Experiment {experiment.id} does not have any dataframes logged to it."
                 )
@@ -123,12 +123,8 @@ class DataframePlot(VizBase):
             self.plotting_func_kwargs["color_discrete_sequence"] = get_rubicon_colorscale(
                 len(self.experiments),
             )
-        try:
-            if self.data_df.empty:
-                raise Exception(f"No dataframe with name {self.dataframe_name} found!")
-        except:
-            if self.data_df == None:
-                raise Exception(f"No dataframe with name {self.dataframe_name} found!")
+        if self.data_df is None:
+            raise Exception(f"No dataframe with name {self.dataframe_name} found!")
 
     def register_callbacks(self, link_experiment_table=False):
         outputs = [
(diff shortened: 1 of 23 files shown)


rubicon_ml/viz/dataframe_plot.py
```diff
@@ -89,8 +89,15 @@ def load_experiment_data(self):
         `dataframe_name` must have the same schema.
         """
         self.data_df = None
+        count = 0
 
         for experiment in self.experiments:
+            if len(experiment.dataframes()) == 0:
+                print(
```
ryanSoley, line 96: import warnings and use `warnings.warn` instead. then you won't need the "WARNING: " text either in the string output

rubicon_ml/viz/dataframe_plot.py
```diff
@@ -89,8 +89,15 @@ def load_experiment_data(self):
...
+                    f"WARNING: Experiment {experiment.name} does not have any dataframes logged to it. "
```
ryanSoley, line 97: experiments are generally referenced by ID instead of name as name is not required to be unique

rubicon_ml/viz/dataframe_plot.py
```diff
@@ -89,8 +89,15 @@ def load_experiment_data(self):
...
+                )
+                continue
+            count += 1
             dataframe = experiment.dataframe(name=self.dataframe_name)
```
ryanSoley, line 101: we also want warn if there is a dataframe, but it isn't named `dataframe_name`. right now, we'd still get an error here if there is no dataframe named `dataframe_name`

instead of checking if `len(experiment.dataframes()) == 0` and warning, we should wrap this getter in a try/except and warn instead of erroring in the except block

stephenpardy (commented):
Is it possible to add a test here?

ryanSoley (commented):
let's do this slightly differently, leveraging try/except on the getter rather than just checking if there are no dataframes

rubicon_ml/viz/dataframe_plot.py
```diff
@@ -116,6 +123,9 @@ def load_experiment_data(self):
                 len(self.experiments),
             )
 
+        if count == 0:
+            raise Exception(f"No dataframe with name {self.dataframe_name} found!")
```
stephenpardy, line 127: Also we can check if `self.data_df` is `None` here instead of the counter.

jeh362:
> Is it possible to add a test here?

I wasn't exactly sure how to write a test for this and tested using the notebooks but I can look into another testing strategy.

stephenpardy:
> > Is it possible to add a test here?
> 
> I wasn't exactly sure how to write a test for this and tested using the notebooks but I can look into another testing strategy.

Check out https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions and you can make mocked situations that have/don't have the dataframes to see that it raises at the right time.

jeh362:
> > > Is it possible to add a test here?
> > 
> > 
> > I wasn't exactly sure how to write a test for this and tested using the notebooks but I can look into another testing strategy.
> 
> Check out https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions and you can make mocked situations that have/don't have the dataframes to see that it raises at the right time.

Thank you I will look into this!

rubicon_ml/viz/dataframe_plot.py
```diff
@@ -91,7 +93,13 @@ def load_experiment_data(self):
         self.data_df = None
 
         for experiment in self.experiments:
-            dataframe = experiment.dataframe(name=self.dataframe_name)
+            try:
+                dataframe = experiment.dataframe(name=self.dataframe_name)
+            except Exception:
```
stephenpardy, line 98: Can we limit this to only the relevant exception?

rubicon_ml/viz/dataframe_plot.py
```diff
@@ -115,6 +124,8 @@ def load_experiment_data(self):
             self.plotting_func_kwargs["color_discrete_sequence"] = get_rubicon_colorscale(
                 len(self.experiments),
             )
+        if self.data_df is None:
+            raise Exception(f"No dataframe with name {self.dataframe_name} found!")
```
stephenpardy, line 128: Should this be RubiconException?

tests/unit/viz/test_dataframe_plot.py
```diff
@@ -74,3 +74,12 @@ def test_dataframe_plot_register_callbacks_link(viz_experiments, is_linked, expe
...
     assert registered_callback_name == "update_dataframe_plot"
     assert registered_callback_len_input == expected
+
+
+def test_dataframe_not_logged_all_experiments(viz_experiments):
+    for exp in viz_experiments:
+        if len(exp.dataframes()) == 0:
+            viz_experiments.remove(exp)
+    dataframe_plot = DataframePlot("test dataframe", experiments=viz_experiments)
+    with pytest.raises(Exception):
```
stephenpardy, line 84: We should check here for whatever we raise above

rubicon_ml/viz/dataframe_plot.py
jeh362: I am not sure I just tried this and ran into some errors with the testing.

stephenpardy: See my comment below - we will need to test for the same exception we raise here. The typical pattern is to raise one of the [concrete exceptions](https://docs.python.org/3/library/exceptions.html#concrete-exceptions) or an exception that we control (like the Rubicon Exception). Since we mostly raise the `RubiconException` throughout I think we should do so here.

jeh362: I tried raising and asserting a `RubiconException` however this is failing the test. I am not sure why because when I also try this in the notebook, the `RubiconException` is being raised.

stephenpardy (approved):

Issue #469, opened 2024-07-25
`DataframePlot` - don't require `dataframe_name`


ryanSoley:
**Is your enhancement request related to a problem? Please describe**

it should be possible to instantiate visualizations without any arguments other than their experiments

**Describe the solution you'd like**

`dataframe_name` should default to None
> https://github.com/capitalone/rubicon-ml/blob/main/rubicon_ml/viz/dataframe_plot.py#L46

if no `dataframe_name` is passed in, assume the dataframe requested is the first dataframe in the first experiment, e.g. `self.dataframe_name = self.experiments[0].dataframes()[0].name`. if `experiments[0]` has no dataframes, raise an error

PR #472, opened 2024-07-25
Dataframe name

jeh362:
closes: #469

---
**Issue**
An error is raised whenever a dataframe is created without a `dataframe_name` parameter. Visualizations should be able to be instantiated without any arguments other than their experiments.

**Changes**
Updated `dataframe_plot.py` to default `dataframe_name` to `None`. If no `dataframe_name` is passed in, then the `dataframe_name` is set to the name of the dataframe logged to the the first experiment.

### DIFF
--- a/rubicon_ml/viz/dataframe_plot.py
+++ b/rubicon_ml/viz/dataframe_plot.py
@@ -92,8 +92,10 @@ class DataframePlot(VizBase):
         count = 0
 
         for experiment in self.experiments:
-            if (len(experiment.dataframes())==0) :
-                print(f"WARNING: Experiment {experiment.name} does not have any dataframes logged to it. ")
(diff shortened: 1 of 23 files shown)


rubicon_ml/viz/dataframe_plot.py
```diff
@@ -43,17 +43,22 @@ class DataframePlot(VizBase):
 
     def __init__(
         self,
-        dataframe_name,
+        dataframe_name=None,
```
stephenpardy, line 46: Can we please add type hints as we go through these? Thanks.

stephenpardy:
Is this PR dependent on #471 ?

ryanSoley:
will address later if necessary

PR #475, opened 2024-08-15
experiment dataclass - warn on extra kwargs

ryanSoley:
followup to #474

## What
  * overrides the domain `Experiment`'s `dataclass`-created `__init__`
    * warn on unexpected kwargs rather than error to support forwards and backwards compatibility across changes to the rubicon-ml data model
    * from this point forward, any rubicon-ml library version should be able to read experiments from disk written by any other rubicon-ml library version (assuming only the data model changes, not the way the library itself reads data)
  * moves shared initialization code between `Project` and `Experiment` into `InitMixin`

## How to Test
  * `python -m pytest tests/unit/domain/test_domain.py`

## Next Steps
  * repeat the process for the remaining domain-level `dataclass`es

### DIFF
--- a/rubicon_ml/domain/experiment.py
+++ b/rubicon_ml/domain/experiment.py
@@ -1,24 +1,86 @@
-from __future__ import annotations
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, List, Optional
 
-from dataclasses import dataclass, field
-from datetime import datetime
-from typing import List, Optional
+from rubicon_ml.domain.mixin import CommentMixin, InitMixin, TagMixin
 
-from rubicon_ml.domain.mixin import CommentMixin, TagMixin
-from rubicon_ml.domain.utils import TrainingMetadata, uuid
+if TYPE_CHECKING:
+    from datetime import datetime
+
+    from rubicon_ml.domain.utils import TrainingMetadata
 
 
-@dataclass
-class Experiment(TagMixin, CommentMixin):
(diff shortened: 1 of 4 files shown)


rubicon_ml/domain/mixin.py
```diff
@@ -1,53 +1,93 @@
...
+            A list of string comments to remove from this domain model.
         """
-        self.tags = list(set(self.tags).difference(set(tags)))
+        self.comments = list(set(self.comments).difference(set(comments)))
 
 
-class CommentMixin:
-    """Adds comment support to a domain model."""
+class InitMixin:
+    """Adds helpers for initializing default values."""
```
ryanSoley, line 36: this new class is the only actual change to this file - I swapped the `CommentMixin` and the `TagMixin` to alphabetize the classes

rubicon_ml/domain/experiment.py
```diff
@@ -1,24 +1,87 @@
...
+        self.branch_name = branch_name
+        self.comments = comments or []
+        self.commit_hash = commit_hash
+        self.created_at = self._init_created_at(created_at)
+        self.description = description
+        self.id = self._init_id(id)
+        self.model_name = model_name
+        self.name = name
+        self.tags = tags or []
+        self.training_metadata = training_metadata
```
stephenpardy, line 87: Should there be a warning here on the kwargs?

stephenpardy (approved):

PR #481, opened 2024-08-29
read/write repository regression tests

ryanSoley:
## What
  * adds regression tests to validate all interactions between the repository layer and the filesystem itself
    * `test_read_regression` tests that the repository accurately retrieves data from the filesystem
    * `test_write_regression` tests that the repository accurately writes data to the filesystem
    * `test_delete_regression` tests that the repository accurately removes data from the filesystem
    * `test_read_write_regression` tests that the repository accurately retrieves data written by the repository itself

## How to Test
  * `python -m pytest tests/regression`


tests/regression/test_repository_read_write.py
stephenpardy, line 1: Can we split these tests into either parameterized versions or just smaller ones that test one (or a few) things at a time?

stephenpardy (approved):

PR #483, opened 2024-09-04
deprecate prefect tasks

ryanSoley:
## What

deprecates the `rubicon_ml.workflow` module, only containing `prefect` tasks. rubicon-ml can still be used within custom prefect tasks as it normally would be
  * these tasks have not been updated in over 3 years and are not leveraged by any known users
  * they can be simply recreated, as shown in the update example notebook that will remain in the docs
  * frequent major updates to `prefect` have caused conflicts and vulnerabilities we no longer want to deal with inside rubicon-ml by having prefect as a dependency

## How to Test
  * `python -m pytest`


rubicon_ml/workflow/prefect/__init__.py
```diff
@@ -8,6 +11,13 @@ def _check_for_prefect_extras():
         raise ImportError(message)
 
 
+warnings.warn(
+    "The `rubicon_ml.workflow.prefect` module is deprecated and will be removed in an upcoming release."
```
shania-m, line 15: do we need it twice

rubicon_ml/workflow/__init__.py
```diff
@@ -0,0 +1,8 @@
+import warnings
+
+warnings.warn(
```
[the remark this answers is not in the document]
shania-m: lets get a test to make sure the warnings are raised

rubicon_ml/workflow/prefect/__init__.py
ryanSoley: yeah, I wanna make sure its raised if the workflow module is imported at any level since we're deprecating all of it. removing the additional warnings makes some of the new tests fail

shania-m: oh wow, I thought having it in the init would cover all the cases . Good to know

shania-m (approved):
LGTM

PR #484, opened 2024-09-19
pass s3 storage options to dataframe read/write

ryanSoley:
## What
  * `storage_options` were only being passed to `fsspec` for standard file read/write
  * now passes provided storage options to dataframe read/write functions where necessary
    * `storage_options` must be empty for non-remote storage or pandas and dask will error

## How to Test
  * `python -m pytest tests`

### DIFF
--- a/rubicon_ml/repository/base.py
+++ b/rubicon_ml/repository/base.py
@@ -43,7 +43,7 @@ class BaseRepository:
     """
 
     def __init__(self, root_dir: str, **storage_options):
-        self._df_storage_options = {}
+        self._df_storage_options = {}  # should only be non-empty for S3 logging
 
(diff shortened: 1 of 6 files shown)


rubicon_ml/repository/base.py
```diff
@@ -614,7 +616,7 @@ def _persist_dataframe(
             df.write_parquet(path)
```
ryanSoley, line 616: polars only seems to offer storage option input on read, not write
> https://docs.pola.rs/api/python/stable/reference/api/polars.DataFrame.write_parquet.html

dianelee217 (approved):

thebrianbn (approved):
LGTM

PR #485, opened 2024-09-20
remove prefect dependency

ryanSoley:
## What
  * next step in deprecating `workflow` module - followup to #483
  * removes the hard dependency on `prefect`
    * `workflow` module is still usable if `prefect` is installed explicitly

## How to Test
  * `python -m pytest`


thebrianbn (approved):
LGTM

shania-m (approved):
LGTM

PR #512, opened 2025-04-24
test: use fixtures, clean up paths & pin dask

ryanSoley:
## What
  * uses the fixtures with the proper clean up functionality rather than instantiating `Rubicon` objects directly in tests
    * not sure how these tests weren't interfering with each other in their current state
  * remove relative paths from the memory filesystem
    * the memory filesystem has no concept of a relative path since there is no location relative to an irl pwd
  * pins dask to avoid using 2025.4.0 released on Tuesday (4/22/25)
    * causing errors with dask-xgboost: `TypeError: Value type is not supported for data iterator:<class 'distributed.client.Future'>`
    * this error was not introduced in this PR - the last few edgetest runs have been failing as well

## How to Test
  * `python -m pytest`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,7 @@ s3 =
 	s3fs<=2025.3.2,>=0.4
 test = 
 	backports.strenum<=1.3.1,>=1.3.1; python_version < "3.11"
-	dask[dataframe,distributed]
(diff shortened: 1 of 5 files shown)


thebrianbn (approved):
LGTM

PR #515, opened 2025-05-07
feat!: remove prefect

ryanSoley:
## What
  * removes the prefect modules that were deprecated in #483 
    * deprecation messages have been in the module since September 2024
    * prefect 2.x can no longer run with the most recent versions of the other libraries in rubicon-ml's dependencies

## How to Test
  * `python -m pytest`
  * validate that edgetest runs again
    * https://github.com/capitalone/rubicon-ml/actions/runs/14888141081

PR #517, opened 2025-05-09
chore: update `setuptools` & add python 3.12 tests

ryanSoley:
## What

  * updates `setuptools` to use at least the current latest version (80.3.1)
    * the action has been using a version in the mid 60s, triggering deprecation emails from PyPi every time we release
  * adds 3.12 to the test matrix
    * allows use of latest `setuptools`
    * was blocked by prefect stuff removed in #515 
  * only validates build in 3.12 during tests
    * the goal of the build validation step is to ensure the publish workflow will execute properly off main during a release
      * the publish workflow only uses 3.12, so only testing 3.12 should be sufficient

## How to Test

```bash
conda create -n rubicon-ml-build python=3.12
conda activate rubicon-ml-build
pip install -e './[build]'
python setup.py sdist bdist_wheel
pip install dist/*.whl
python -c 'import rubicon_ml; print(rubicon_ml.__version__)'
```

the last command should output something like this: `0.10.5+2.g7269548`

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -30,7 +30,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install -e './[all,qa,test]'
     - name: Install build dependencies
-      if: ${{ matrix.python-version }} == '3.12'
+      if: ${{ matrix.python-version == '3.12' }}
       run: |
         pip install -e './[build]'
     - name: Check formatting with black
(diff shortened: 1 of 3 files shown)


.github/workflows/test-package.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     - name: Validate build
       run: |
         python setup.py sdist bdist_wheel
-        pip install dist/rubicon_ml-*.whl
+        pip install dist/rubicon*ml-*.whl
```
ryanSoley, line 44: newer versions of setuptools use the package name with an underscore, where old versions used the library name with a dash. allows for both so we can test build in all available versions

stephenpardy (approved):

PR #541, opened 2025-07-25
feat: `h2o` schema updates

ryanSoley:
**To switch to the release PR template go to the `Preview` tab, [click here](?template=release.md), then click `Create pull request` on the next page.**

---

## Changes
  * adds an `h2o` base model schema that each `h2o` schema now extends
    * logs the `actual_params` and `model_id` attributes as parameters
    * logs `auc`, `aucpr`, and `logloss` as metrics

## How to Test
  * `uv run pytest tests/integration/test_schema.py `


.github/workflows/publish-package.yml
```diff
@@ -17,6 +17,8 @@ jobs:
   deploy:
     if: github.event.pull_request.merged == true && startsWith(github.head_ref, 'release/')
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
ryanSoley, line 21: this should fix the issue with the last release not being able to tag the repo with the new version

PR #561, opened 2025-11-10
feat: add `wandb` backend

ryanSoley:
## Changes
  * adds `WandBRepository` to enable rubicon-ml logging to and retrieval from [Weights & Biases](https://wandb.ai/site)
    * updates existing repository-level regression tests for new backend
    * adds client-level regression tests for new and existing backends
  * updates JSON serialization to adhere to new datetime standards
    * preserves and tests backwards compatibility of existing datetime logs

## How to Test
  * once locally authenticated with `wandb`, instantiate a `Rubicon` object with "wandb" `persistence`
    * `rubicon = Rubicon(persistence="wandb", entity="my wandb team")`
      * `entity` is optional and defaults to the logged in user's personal space
      * all existing rubicon-ml functionality should work from this `rubicon` entry point
    * change the `persistence` in the example notebooks and run them to validate
  * `python -m pytest tests/regression/test_client_read_write.py -m wandb`
  * `python -m pytest tests/regression/test_repository_read_write.py -m wandb`


tests/regression/test_client_read_write.py
```diff
@@ -0,0 +1,372 @@
...
+        )
+
+    with temp_dir_context as temp_dir_name:
+        root_dir = os.path.join(temp_dir_name, "test-rubicon-ml")
+        rubicon = Rubicon(persistence=persistence, root_dir=root_dir)
+        project = rubicon.get_or_create_project(**project_parameters)
+        experiment = project.log_experiment(**experiment_parameters)
+
+        if is_existing_experiment and persistence == "wandb":
+            rubicon.repository._active_run.finish()
```
nagireddyakshay: Instead of accessing a private variable, lets expose a public method something like `finish_run` in the `wandb` module which would do something below

```
if self._active_run is not None:
    self._active_run.finish()
    self._active_run = None
```

We can then expose a `finish()` method in the `experiment` module for the end users =

Anticipated usage
```
if persistance == "wandb""
    experiment.finish()
retieved = experiment.metric(name="acc")
```

rubicon_ml/client/mixin.py
```diff
@@ -777,9 +778,10 @@ def _get_commentable_identifiers(self):
         project_name, experiment_id = self._parent._get_identifiers()
         entity_identifier = None
 
-        # experiments do not return an entity identifier - they are the entity
+        # experiments are identified by their `id`s
```
mirikerstein: Looks like this bit of logic is duplicated

rubicon_ml/repository/wandb.py
```diff
@@ -0,0 +1,1314 @@
...
+            )
+
+        self.root_dir = root_dir
+
+        self.entity = entity
+        self.storage_options = storage_options
+
+        self._active_run = None
+        self._current_artifact_bytes = None
+        self._current_dataframe = None
```
nagireddyakshay: I noticed these two instance variables are being used to pass data b/w  `_persist_bytes()`, `_persist_dataframe()` and `_persist_domain()`, and I believe this could create issues in a multi-thread context - Multiple concurrent calls would overwrite the shared instance state. (Very unlikely scenario someone uses this in a multi-thread context thought)

Lets instead complete override the `create_artifact()` and `create_dataframe()` methods insted of using the base class pattern, something like below


```
def create_artifact(self, artifact, data, ...):
      if experiment_id is None:
         raise ...
       
      with tempfile.TemporaryDirectory() as temp_dir:
              temp_path = os.path.join(temp_dir, artifact.name)
              with open(temp_path, 'wb') as f:
                      f.write(data)
               wandb_artifact = self.wandb.Artifact(name=artifact.name, type="model')
               artifact.add_file(temp_path, name=entity.name)
               artifact.save()
                
                entity_id = entity.id
                entity.id = f"{entity.id}:{entity.name}"  # for accurate W&B retrieval

                self._set_config_value(f"_rubicon_artifact_{entity_id}", json.dumps(entity))

````

We can do the same as above with `create_dataframe()` as well and then remove the `_persist_bytes/_persist_dataframe` methods and the respective instance variables entirely.

File: rubicon_ml/repository/wandb.py [hunk 4]
```diff
@@ -0,0 +1,1314 @@
...
+
+        if result is None:
+            raise RubiconException(
+                f"No experiment metadata found for experiment '{experiment_id}'. "
+                "This run was not created with `rubicon-ml`."
+            )
+
+        return result
+
+    def get_experiments(self, project_name: str) -> List[domain.Experiment]:
```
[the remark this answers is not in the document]
nagireddyakshay: Lets also add `filter` param to this for enabling server-side filtering. [Ref](https://docs.wandb.ai/models/ref/python/public-api/runs#method-runs-__init__)

ryanSoley:
> Can we please update the Tutorials and all other relevant sections in the docs with the wandb references?

I've added references to the new W&B backend to the API spec where relevant. there are not currently documentation pages for specific backend implementations, so I've got to think about how we want to add them

ryanSoley:
> LGTM Just a thought, not requesting changes on this, but I don't love that we're exposing a finish method and it's not available for all backends...

@mirikerstein yeah I thought the same thing at first but didn't really come up with anything better - did you have any thoughts for another approach? I don't think its a huge deal for it to just exist as a no-op on the repos that don't support it, and it's entirely possible that other future backends could need it

shania-m (approved):

PR #588, opened 2026-04-14
feat: deprecate intake

ryanSoley:
## Changes
  * remove `intake` from core dependencies
    * the latest versions of `intake` are vulnerable, causing whitesource to block PRs, and there doesn't appear to be much momentum on the library to fix it
  *  deprecate core functionality that uses `intake` (`publish`)
  * removes 'binder/requirements.txt'
    * it pulls in literal rubicon, so since it exists with the intake vulnerability there's no solving without a new release
      * we can put it back after we cut a new release or just totally deprecate binder - idt it works even with this file 

## How to Test
  * `python -m pytest`

### DIFF
--- a/rubicon_ml/intake_rubicon/__init__.py
+++ b/rubicon_ml/intake_rubicon/__init__.py
@@ -1,17 +1,42 @@
-import intake  # noqa F401
+import importlib
+import logging
+import warnings
 
-from rubicon_ml.intake_rubicon.experiment import ExperimentSource
-from rubicon_ml.intake_rubicon.viz import (
-    DataframePlotDataSource,
-    ExperimentsTableDataSource,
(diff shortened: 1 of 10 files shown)


shania-m (approved):

rubicon_ml/intake_rubicon/__init__.py
```diff
@@ -1,17 +1,42 @@
...
 
-__all__ = [
-    "ExperimentSource",
-    "ExperimentsTableDataSource",
-    "MetricCorrelationPlotDataSource",
-    "DataframePlotDataSource",
-    "MetricListComparisonDataSource",
-]
+
+def _check_intake():
```
shania-m, line 14: Can we get a paramatrized unit test for this function?

PR #592, opened 2026-04-30
fix: `wandb` entity

ryanSoley:
## Changes
  * actually references the input `entity` when creating run configurations for `wandb` team logging

## How to Test
  * run the wandb regression test


armaan-dhillon (approved):
Lgtm

PR #602, opened 2026-06-18
fix: composite config in project client

ryanSoley:
## Changes

fixes a composite config bug where `get_or_create_project` silently dropped secondary backends when only some, but not all backends had an existing project with the requested name

## How to Test

  * `python -m pytest tests/unit/client/test_rubicon_client.py`
  * validate the following MRE now succeeds:
```python
import uuid                            

composite_config = [
    {"persistence": "filesystem", "root_dir": "./rubicon-data-a"},
    {"persistence": "filesystem", "root_dir": "./rubicon-data-b"},
]
project_name = str(uuid.uuid4())

# simulate project pre-existing in backend A only (prior run, non-composite client, etc.)
rubicon_a = Rubicon(**composite_config[0])
rubicon_a.create_project(project_name)

# composite client finds it in A, takes get path, returns single-config project
rubicon = Rubicon(composite_config=composite_config)
project = rubicon.get_or_create_project(project_name)
experiment = project.log_experiment()

rubicon_b = Rubicon(**composite_config[1])
rubicon_b.get_project(project_name)  # originally fails as project never created in B
```

### DIFF
--- a/rubicon_ml/client/rubicon.py
+++ b/rubicon_ml/client/rubicon.py
@@ -202,7 +202,7 @@ class Rubicon:
                 except Exception as err:
                     return_err = err
                 else:
-                    return Project(project, self.config)
+                    return Project(project, self.configs)
 
             if len(self.repositories) > 1:
                 raise RubiconException("all configured storage backends failed") from return_err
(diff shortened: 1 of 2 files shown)


armaan-dhillon (approved):
LGTM

